PR #60, opened 2026-07-20
chore: update .dev.vars.example with actual config values (account ID…

AMM2034567:
…, RTK app ID, AI gateway URL)

## Testing

- [ ] Tested locally
- [ ] Checked on production / staging


AMM2034567:
1

PR #95, opened 2026-08-23
[codex] add room-isolated meeting transcript memory and Doubao PCM ingress

madawei2699:
## Summary

This PR connects the production Meeting Notes STT path to the resident Agent Harness and fixes the raw RTP Opus ingress used by the Doubao provider.

Previously, the runtime could receive and transcribe audio in a canary, but the real `AgentDaemon` discarded STT events before a Harness turn. A resident Agent therefore could not answer a later text `@` question such as “what did I just say?”. Cloudflare also delivered raw RTP Opus packets while the provider request treated them as an Ogg stream, so Doubao rejected the audio format.

## Changes

- Decode raw RTP Opus locally with `opus-decoder`.
- Convert decoded stereo audio to bounded 16 kHz mono PCM before sending it to Doubao.
- Keep bounded startup headroom for the provider handshake.
- Store only committed STT text in a temporary, private transcript file.
- Isolate transcript state per room join under:
  `~/.free4chat-agent/workspaces/<instance-uuid>/.meeting-notes/transcript.jsonl`
- Keep `participantId` and `participantName` on every segment for multi-speaker attribution.
- Pass a bounded transcript snapshot and the workspace-relative file capability `.meeting-notes/transcript.jsonl` to addressed Harness turns.
- Treat transcript initialization and write failures as fail-open: text Agent joining and `@Agent` turns continue using the in-memory snapshot.
- Clean transcript/runtime resources on room expiry and remove stale per-instance workspaces when a new daemon starts after a crash.
- Do not ship the temporary canary persistence script in the product package.

Raw audio, partial results, credentials, participant handles, and transcript state are not sent to the Worker/DO and are not persisted in R2, KV, or Durable Object storage. TTS and outbound voice remain out of scope.

## Isolation and safety

The daemon creates a fresh UUID workspace for every room join. A runtime instance has one fixed room ID and its Harness process uses that instance workspace, so joining another room creates a separate runtime, Harness, and transcript. The Harness may read only the explicitly supplied workspace-relative transcript file; other workspace files, shell commands, MCP calls, credentials, and external services remain prohibited.

## Validation

- `npm test` — 77/77 passed
- `npm run type-check` — passed
- `npm run lint` — passed
- `npm run build` — passed
- `git diff --check` — passed
- Local production canary with the configured Doubao key verified continuous RTP frames, partial results, committed results, and Stop → Start recovery. The key was not printed or committed.

## Follow-up

After merge, validate a real resident Harness E2E with two or more humans speaking, then ask the Agent text questions about specific speakers and facts. The next Meeting Notes work is bounded 30+ minute context and notes output (topics, decisions, action items, open questions). TTS/realtime voice reply remains the later Voice Agent phase.


madawei2699:
Addressed the review changes in commit `dc97d30`:

- Transcript flush failures are fail-open: the in-memory snapshot still reaches the Harness and the addressed text turn is not consumed. Added regression coverage.
- Natural room expiry now closes Runtime resources, removes the instance from the daemon, and deletes its workspace. A fresh daemon removes stale per-instance workspaces after the previous daemon is unreachable. Added regression coverage.
- Removed the repository canary script that persisted transcript output under `/tmp`.
- Harness transcript path is now workspace-relative (`.meeting-notes/transcript.jsonl`), so the local home path is not included in the prompt.

Local validation: 76/76 tests, type-check, lint, build. GitHub Actions: Agent Runtime checks passed.

madawei2699:
Addressed the final blocking review item in commit `08e874b`:

- `MeetingTranscriptStore.ready()` is now fail-open in `ResidentRoomRuntime.start()`. Initialization errors emit only `meeting_transcript_init_failed`; the runtime still joins and addressed text turns still reach the Harness.
- Added a regression test covering init failure, reply delivery, and sanitized logging.
- Updated the PR body to describe the workspace-relative transcript capability.

Validation: 77/77 tests passed locally, type-check/lint/build passed, and both Agent Runtime CI checks are green.

PR #99, opened 2026-08-23
[codex] wait for SFU connection before subscriptions

madawei2699:
## Problem

The resident Meeting Notes Runtime completed its initial DataChannel SDP exchange and immediately listed/subscribed to Human tracks. Cloudflare Realtime requires the WebRTC PeerConnection to reach `connected` before remote-track operations; otherwise those operations can block or time out. In a real room this left Meeting Notes in `starting` even though the UI showed Listening.

## Fix

- Build the same initial, subscribe-only transport shape as the browser: one recvonly audio transceiver plus the `server-events` DataChannel before the first offer.
- Wait for werift ICE/DTLS `connected` after the initial SDP exchange and before `agent-room-media` or `/tracks`; the wait is bounded to 10 seconds and failure remains fail-open for text Agent operation.
- Surface a sanitized `meetingNotesMediaState` (`idle`, `starting`, or `running`) in local Runtime status so field diagnostics do not need tokens, SDP, raw audio, or transcript contents.
- Preserve only the safe SFU route/error-code diagnostic already used by the Runtime; upstream response bodies remain hidden.

## Validation

- `npm run format`
- `npm test` — 85 passing tests
- `npm run lint`
- `npm run type-check`
- `npm run build`
- Direct real-SFU probe with the local credential: agent session and DataChannel establish both returned 200; a second probe confirmed werift accepts the returned SDP answer.
- Local Worker-route integration probe using the actual `handleSfuRequest` path and real SFU also returned 200 for agent session and DataChannel establish.

No Worker/DO authorization, transcript behavior, STT provider behavior, speaker attribution, or media publishing capability changes are included.

### DIFF
--- a/agent-runtime/src/core/runtime.ts
+++ b/agent-runtime/src/core/runtime.ts
@@ -35,6 +35,9 @@ export interface RuntimeStatus {
   state: "starting" | "waiting" | "turn" | "reconnecting" | "stopped" | "failed"
   participantId?: string
   lastError?: string
+  /** Optional Meeting Notes bridge lifecycle state. This is deliberately
+   * status-only: no media credentials, SDP, audio, or transcript content. */
+  meetingNotesMediaState?: "idle" | "starting" | "running"
 }
 
 export interface ResidentRuntimeOptions {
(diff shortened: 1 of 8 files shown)


madawei2699:
Closing: this was an early werift-based transport experiment. Subsequent live investigation showed werift fails at the initial Cloudflare DataChannel SDP exchange, so waiting before subscriptions cannot make it production-ready. The retained evidence and the transport-neutral local debugging runbook are in #101; future engine evaluation is tracked in #100.

Issue #165, opened 2026-08-28
protocol: structured Agent text addressing and conversational handoff


madawei2699:
## Summary

Fix the current Human→Agent / Agent→Agent text-addressing asymmetry so independently running Agents can explicitly hand a conversation to one another without relying on natural-language `@Name` parsing.

Dogfooding a multi-Agent Room exposed the failure clearly:

```text
Human → @Pi-Agent
Pi replies: "@Hermes Agent @Codex @OpenCode Agent 接上..."
```

The text is visible in the Room, but Hermes/Codex/OpenCode do not wake.

That is expected from the current implementation:

```text
Human browser
  text + targets[]
      ↓
RoomMessage.targets[]
      ↓
addressed=true for target Agent(s)
      ↓
resident Runtime wakes Harness
```

while Agent text currently follows:

```text
HarnessTurnResult { Text }
      ↓
Runtime SendText(handle, text)
      ↓
MCP send_text(participantHandle, text)
      ↓
agent-send-text { text }
      ↓
RoomMessage with NO targets
```

So an Agent-generated literal `@Hermes` is only prose. It has no activation semantics.

This is a #155 dogfood blocker for natural multi-Agent collaboration and a protocol-symmetry follow-up to #106.

## Core principle

> **Visible `@Agent` is a human-readable representation of addressing. Structured target metadata is the source of truth.**

Do not make routing depend on regex/NLP parsing of message text.

A sentence such as:

```text
I agree with what @Hermes said earlier.
```

must not accidentally wake Hermes merely because a name appears in prose.

The Room should know who is being activated because the sender supplied explicit target participant IDs.

## Collaboration semantics

Preserve the current distinction:

```text
Visibility ≠ Activation
```

### Visibility

Ordinary Room events remain shared Room context. All connected participants may observe them through the existing bounded event stream.

### Activation

A resident Agent Harness receives a new turn only when an event is explicitly addressed to that Agent.

Conceptually:

```text
unaddressed message
    → all participants may observe it as Room context
    → no new Harness turn

addressed message
    → all participants may observe it as Room context
    → only target Agent(s) get a Harness turn
```

This behavior is important in multi-Agent Rooms: waking every Agent for every message can create chatter loops, token waste, duplicate responses, and unstable turn ownership.

## Existing protocol foundation

Do not invent a second addressing model.

The Human browser path already has the right primitive:

```ts
{ type: "chat", text: string, targets?: string[] }
```

[...]

Issue #170, opened 2026-08-28
media: make Agent voice a per-participant capability with independent mute/unmute


madawei2699:
## Status

**DOGFOOD FINDING from #155 — previous speaker-picker direction superseded.**

Real multi-Agent use showed that the existing Room-wide `voiceReply.agentParticipantId` model is the wrong interaction abstraction.

The current UI silently chooses the first Agent in roster order. The first proposed fix for this issue was to add a speaker picker. Further dogfood with several Agents on one machine showed that a picker would only make the wrong abstraction explicit:

```text
one Runtime Host
├─ Pi
├─ Codex
├─ Hermes
└─ one local Doubao credential / TTS capability
```

The credential and speech engine belong to the local Free4Chat Runtime Host, not to one Harness/Agent participant. If the Runtime can synthesize speech, every Agent resident on that Runtime can potentially speak.

Therefore Voice should not mean:

```text
Room Voice Reply
→ choose exactly one Agent
```

It should mean:

```text
Runtime Host has TTS capability
→ that capability is projected onto its Agent participants
→ each eligible Agent has an independent Room-visible mute/unmute state
→ multiple Agents may be voice-enabled at the same time
```

## Product model

Voice is participant presentation, because speech output belongs to a specific speaker.

Example:

```text
Runtime A — TTS ready
├─ Pi       🔇
├─ Codex    🔇
└─ Hermes   🔇

Runtime B — TTS unavailable
└─ Claude   voice unavailable
```

A Human may independently enable Pi and Codex:

```text
Pi       🔊
Codex    🔊
Hermes   🔇
Claude   —
```

Then:

```text
@Pi ...
→ Pi replies
→ Runtime A TTS
→ Pi voice is published to Room

@Codex ...
→ Codex replies
→ Runtime A TTS
→ Codex voice is published to Room
```

No Human should have to stop Voice, choose another Agent, and restart Voice merely to hear a different Agent on the same Runtime.

## Capability ownership

The important boundary is:

> **Credentials belong to the Runtime Host. TTS capability is provided by the Runtime Host and projected onto its resident Agent participants. Voice enable/mute remains participant-specific.**

Room/UI must never receive provider keys or credential details.

The Room only needs a safe readiness projection such as:

```text
voiceAvailable = true | false
```

or equivalent host capability metadata from the foundation issue.

If a Runtime does not have valid TTS capability, its Agents should show a disabled/grey voice control or no voice control at all.

[...]

PR #175, opened 2026-08-28
fix(agent): report speech prerequisites for the active grant

### LINKED ISSUE #171: runtime: make speech prerequisite notices truthful on the current media path
## Status

**CURRENT correctness bug on the shipped participant-bound media path.**

Dogfood under #155 proved that enabling Voice Reply can cause a resident Agent to post:

> Meeting Notes was requested, but no speech-to-text provider is configured...

when Meeting Notes was never requested.

PR #175 currently implements the narrow fix.

Separately, later dogfood refined the product/media model: #170 now moves Agent voice toward Runtime-host TTS readiness + per-Agent mute/unmute, while #177 replaces the special Meeting Notes Agent model with Room-wide Live Transcript. Those follow-ups supersede the long-term single-Agent grant abstraction, but they do **not** make the current false Room message acceptable while the old path remains deployed.

## Current root cause

The shipped Runtime shares one Pion media bridge between the current Meeting Notes and Voice Reply grants.

A grant activation callback is used for both paths, but the old notice builder checks only STT readiness and emits Meeting Notes-specific copy.

Conceptually:

```text
Voice Reply activates
→ shared media bridge activates
→ generic prerequisite callback
→ checks STT
→ falsely says Meeting Notes was requested
```

## Required invariant

> **The Runtime must never claim that a Human requested a capability that was not actually activated.**

On the current implementation path:

```text
Meeting Notes activation
→ evaluate STT prerequisite

Voice Reply activation
→ evaluate TTS prerequisite
```

If both current grants are active, evaluate each independently without poll spam.

## Scope of the fix

Keep this a narrow correctness/stabilization change.

- Voice-only activation must never emit Meeting Notes/STT copy.
- Notes-only activation must never emit Voice/TTS copy.
- Missing prerequisite notice should be edge/grant-epoch driven rather than repeated on every media poll.
- Text collaboration must remain functional when speech is unavailable.
- #167 credential provisioning/hot reload remains intact.
- credential values/details never enter Room messages/logs/ACP/Harness context.

Do **not** use this issue to add more UX around the legacy single-Agent grants.

## Relationship to the new model

The current concepts are transitional:

```text
legacy Voice Reply
→ one agentParticipantId

legacy Meeting Notes
→ one agentParticipantId
```

The dogfood-derived target is now:

```text
#176 Runtime Host capability foundation

TTS
→ Runtime Host readiness
→ per-Agent independent voice mute/unmute (#170)

STT
→ Human-provided STT-ready Runtime
→ one Room-wide Live Transcript producer (#177)
→ transcript becomes shared Room context (#157)
```

Therefore the implementation for this issue must avoid creating new dependencies that make `voiceReply.agentParticipantId` or `meetingNotes.agentParticipantId` harder to retire.

If #170/#177 remove the affected legacy notice path before this fix lands, this issue may be closed as superseded after proving the false notice is no longer reachable.

## Acceptance

- [ ] Voice-only activation never emits a Meeting Notes/STT notice.
- [ ] Voice-only activation with missing TTS emits truthful Voice/TTS prerequisite feedback at most once per activation epoch.
- [ ] Notes-only activation never emits a Voice/TTS notice.
- [ ] Notes-only activation with missing STT emits truthful Notes/STT prerequisite feedback at most once per activation epoch.
- [ ] both current grants can be evaluated independently without poll spam;
- [ ] ready prerequisite produces no missing-config notice;
- [ ] reassignment/stop/start/new epoch behaves deterministically;
- [ ] existing text behavior remains available;
- [ ] #167 credential hot reload is unchanged;
- [ ] no new protocol/product dependency entrenches the legacy single-Agent grant model;
- [ ] focused Runtime/media tests cover the above.

## Non-goals

- designing the new Runtime Host identity/capability model (#176);
- implementing Agent voice mute/unmute (#170);
- implementing Room Live Transcript (#177);
- exposing provider readiness secrets to the browser;
- provider expansion;
- realtime S2S (#156);
- broad Room media redesign.

## Related

- #155 — stabilization/dogfood.
- PR #175 — current narrow implementation candidate.
- #176 — Runtime Host speech capability foundation.
- #170 — target Agent voice model.
- #177 — target Live Transcript model.
- #157 — shared ephemeral Room context.
- #167 — credential provisioning/hot reload.

madawei2699:
Closes #171. **Not merged** — opened for review.

## Root cause (code-confirmed)

The media Controller serves Meeting Notes and Voice Reply over ONE shared Pion bridge (by design). But `OnGrantActivated` fired from the bridge-start completion path — which runs for **any** grant that starts the bridge — and `buildSpeechNotice` checked **only `STTEnabled`**, always emitting "Meeting Notes was requested…". A Voice-Reply-only grant could therefore falsely claim Meeting Notes was requested.

## Fix (no bridge split)

- **Grant-edge model in the controller poll** (#171): each grant kind announces exactly once per grant **instance** (kind + `StartedAt` epoch):
  - unchanged grant across 5s polls → no re-fire (no spam);
  - stop/start or re-grant → fresh epoch → fresh edge;
  - reassignment A → B → A re-arms silently, B's own runtime announces its own edge;
  - a second grant added while the shared bridge is already running still gets its own edge — evaluated independently **without splitting the bridge**.
- Malformed/partial `room_info` (missing epoch) can never re-trigger an already-announced grant.
- `OnGrantActivated` now carries `GrantKind` (`meeting_notes` | `voice_reply`); `buildSpeechNotice(config, kind)` evaluates exactly that grant's prerequisite:
  - Meeting Notes → STT only: *"Meeting Notes was requested, but no speech-to-text provider is configured…"*
  - Voice Reply → TTS only: *"Voice Reply was requested, but no text-to-speech provider is configured…"*
  - both grants → both evaluated independently.
- Wording stays concise and keeps the **no-API-keys-in-the-room** guidance. Notices ride the ordinary **unaddressed** send path (#165, nil targets). #167 credential provisioning / hot reload untouched (a reloaded config is picked up by the next grant edge); credential values never leave the Runtime boundary; text remains fully functional without speech.

## Tests

Controller (`controller_grantnotice_test.go`, deterministic synchronous polls):
1. edges fire once per grant instance; unchanged grant = no spam;
2. later VR grant on the running shared bridge gets its own edge;
3. both grants on one poll → both edges, MN then VR;
4. stop/start → fresh edge per kind (new epochs);
5. reassignment A → B: A quiet, B announces.

Runtime (`speech_notice_test.go`):
6. `buildSpeechNotice` table — voice-only TTS missing → voice notice; voice-only TTS ready (STT missing) → silent **and never mentions Meeting Notes**; notes-only STT missing → notes notice; notes-only STT ready (TTS missing) → silent; keys guidance preserved;
7. `notifySpeechPrerequisite` sends the grant-specific notice via the ordinary unaddressed send (nil targets) and sends nothing when the prerequisite is satisfied.

## Validation

`gofmt` clean, `go vet ./...` clean, `go test ./... -count=1` (11 packages, incl. full media + runtime suites) green, `go build ./cmd/free4chat-agent`, linux + darwin cross-builds, `scripts/check-cleanup.sh` OK.

### DIFF
--- a/agent/internal/media/controller.go
+++ b/agent/internal/media/controller.go
@@ -27,9 +27,11 @@ type ControllerOptions struct {
 	// OnTrackStarted/Ended mirror media bridge events (transcriber wiring).
 	OnTrackStarted func(source speech.AudioSource)
 	OnTrackEnded   func(source speech.AudioSource)
-	// OnGrantActivated fires once per successful grant activation (the
-	// runtime uses it to check speech readiness at grant time).
-	OnGrantActivated func()
+	// OnGrantActivated fires on each grant activation EDGE (#171): a grant
+	// newly targeting this participant, or a fresh epoch of it. The runtime
+	// uses it to evaluate that grant's own speech prerequisite once — never
+	// per poll.
+	OnGrantActivated func(kind GrantKind)
 	// Voice configures outbound Voice Reply (nil = Meeting Notes only).
 	Voice *VoiceConfig
 	// Log receives bounded safe events.
@@ -48,6 +50,16 @@ type VoiceConfig struct {
 	OnSpeakerEvent    func(voice.SpeakerEvent)
 }
 
+// GrantKind identifies which room media grant produced an activation edge.
+// Meeting Notes and Voice Reply are independent grants over one shared
+// media bridge; each carries its own speech prerequisite.
+type GrantKind string
+
+const (
+	GrantMeetingNotes GrantKind = "meeting_notes"
+	GrantVoiceReply   GrantKind = "voice_reply"
+)
+
 // Controller owns the Runtime-side half of the media lifecycle: it polls
 // room_info for the Meeting Notes and voiceReply grants and starts/stops the
 // ONE shared Bridge accordingly. This is the ONLY thing that decides when
@@ -64,12 +76,20 @@ type Controller struct {
 	voiceEpoch         *int64
 	voiceObservedEpoch *int64
 	voiceObservedInit  bool
-	bridge             *Bridge
-	speaker            *voice.Speaker
-	voiceStarting      bool
-	stopped            bool
-	cancel             context.CancelFunc
-	now                func() time.Time
+	// #171 grant-announcement state: the last grant instance (kind + epoch)
+	// whose activation edge was already reported, so an unchanged grant
+	// never re-fires while polls continue. Re-armed when the grant stops
+	// targeting this participant.
+	mnAnnounced      bool
+	mnAnnouncedEpoch *int64
+	vrAnnounced      bool
+	vrAnnouncedEpoch *int64
+	bridge           *Bridge
+	speaker          *voice.Speaker
+	voiceStarting    bool
+	stopped          bool
+	cancel           context.CancelFunc
+	now              func() time.Time
 }
 
 // NewController builds an idle controller.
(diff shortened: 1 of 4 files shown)


madawei2699:
#171 has been updated after additional dogfood, but this PR's **narrow correctness scope remains valid**: fix the false/mismatched prerequisite notice on the currently shipped participant-bound media path.

Do not expand this PR into the new media model. #176 owns Runtime Host capability/readiness; #170 owns per-Agent voice mute/unmute; #177/#157 own Room-wide Live Transcript/shared transcript context.

When reviewing #175, the gate should be: does it fix the current false notice cleanly without making the legacy single-Agent grant model harder to retire?

madawei2699 (commented):
Keep this PR as the narrow transitional correctness fix for #171; do not expand it into #176/#170/#177.

One blocker remains on the current head: a transient `RoomInfo()` error is currently collapsed into `authorized=false` / `vrAuthorized=false`, and the new grant-announcement state then clears `mnAnnounced` / `vrAnnounced`. When the next poll succeeds with the same still-active grant and unchanged `StartedAt`, the controller treats it as a fresh activation edge and emits the prerequisite notice again.

A failed observation must stay distinct from a confirmed grant revocation. Preserve the already-announced epoch across transient `room_info` failures; only re-arm announcement state after a successful observation confirms that the grant no longer targets this participant (or a genuinely new epoch appears).

Please add a focused regression test for: active grant epoch E -> one notice -> transient RoomInfo failure -> same active grant epoch E -> no second notice. Cover both Meeting Notes and Voice Reply or a table/shared helper if that stays simpler.

After that, rerun the existing Go checks. No other architecture work is needed in this PR.

madawei2699:
Blocker fixed in `cf68a5a` (narrow change, 2 files).

**Root cause of the duplicate**: the announcement re-arm (`mnAnnounced/vrAnnounced = false`) ran on every poll that concluded "not authorized" — including the fail-closed path after a transient `RoomInfo` error — so recovery observing the same epoch `E` was treated as a new activation.

**Fix**: re-arm is now gated on a SUCCESSFUL `RoomInfo` observation (`roomInfoObserved`). A transport failure leaves the last announced epoch untouched:

```text
successful active E  -> notice once
transient RoomInfo failure (authorized=false, fail-closed media path unchanged)
successful active E  -> epoch E == announced E -> NO second notice
```

Correct semantics preserved on both sides:
- server **positively reports the grant inactive** during the outage -> successful-inactive observation re-arms; a later genuinely new instance (fresh epoch) announces again;
- grant stopped/started during the outage (new `StartedAt`) -> recovery sees a different epoch -> fresh edge (correct: that IS a new grant instance).

Applied identically to **Meeting Notes and Voice Reply**. Regression tests pin the exact sequence for MN, VR, and both-at-once (`TestControllerTransientRoomInfoFailureDoesNotReannounce*`, `...PreservesBothAnnouncements`). Scope stayed exactly this fix — no #176/#170/#177 work.

Issue #177, opened 2026-08-28
media: replace Meeting Notes with Room-wide Live Transcript


madawei2699:
## Status

**DOGFOOD-DERIVED PRODUCT / MEDIA FOLLOW-UP from #155 and #157.**

Real multi-Agent / multi-device use showed that the current **Meeting Notes → choose one Agent note-taker** model mixes three different concepts:

```text
local STT capability
Room media authorization
shared meeting context
```

The Agent is not actually the note-taker. The native Runtime/Pion/STT pipeline produces transcript text; Agents later interpret/use that text.

The product should therefore move from a special "Meeting Notes Agent" to a Room-wide **Live Transcript** capability.

## Core model

> **STT is provided by one Human-controlled, STT-ready Runtime Host, but committed transcript is Room-shared ephemeral context.**

Conceptually:

```text
Human speech in Room
        ↓
ONE authorized STT-ready Runtime Host
        ↓
streaming STT
        ↓
attributed committed transcript events
        ↓
Room-shared ephemeral context
        ↓
Humans + Agents may consume it
```

No Agent participant owns the transcript merely because its Runtime produced it.

## Why this is not Meeting Notes

The current term suggests:

```text
choose Agent A
→ Agent A listens
→ Agent A owns notes/context
→ Agent A summarizes for others
```

That is the wrong collaboration model for #157.

The intended capability is closer to Zoom-style live transcription:

```text
Live Transcript: Off / On
```

It continuously turns enabled Room speech into attributed text context while active. A summary, decision list, issue update or follow-up remains optional Agent work derived from that shared context.

Useful distinction:

> **Transcription is infrastructure. Interpretation is Agent work.**

## Start ownership

Starting transcription spends/uses a local provider capability and sends Room audio through that Runtime's STT provider.

Therefore when Live Transcript is **Off**, a Human may Start it only through an STT-ready Runtime Host they are authorized to provide/use for this Room (#176).

Example:

```text
Dawei
└─ Runtime A — STT ready ✅

Alice
└─ Runtime B — STT unavailable ❌
```

UI while Off:

```text
Dawei: Live Transcript [Start]
Alice: Live Transcript [Unavailable locally]
```

Dawei clicks Start:

```text
Runtime A becomes the single producer
→ Room Live Transcript becomes globally On
```

Alice did not gain STT capability, but she and her Agents may consume the resulting Room transcript.

## Room-global synchronized state

Live Transcript is one Room state, not one toggle per Human.

[...]

Issue #176, opened 2026-08-28
protocol/runtime: model Runtime Host speech capabilities and Room-scoped ownership


madawei2699:
## Status

**DOGFOOD-DERIVED FOUNDATION.**

Multi-Agent / multi-device dogfood under #155 exposed a missing abstraction between Room participants and the local native Runtime.

Today the Room primarily knows Human and Agent participants. Speech configuration, however, is not truly per Agent:

```text
one native Runtime / daemon
├─ local credential store
├─ STT/TTS provider configuration
├─ Pi resident Agent
├─ Codex resident Agent
└─ Hermes resident Agent
```

The same local Runtime credential/readiness is shared by every resident Agent instance. Treating Voice Reply and Meeting Notes as if one Agent participant individually owns that speech capability creates confusing UI and incorrect authorization/readiness behavior.

We need a deliberately small Runtime Host concept that explains where host-provided capabilities live without turning Free4Chat into a device-management system.

## Core invariant

> **Credentials and host-provided media capabilities belong to the Runtime Host. Agent participants may use/project those capabilities, but the credential itself never becomes participant or Room state.**

A useful ownership split is:

```text
Harness / Agent
→ intelligence, tools, private memory, response content

Runtime Host
→ process lifecycle, local credentials, Pion/media engine, STT/TTS capability

Room
→ temporary participant presence, capability projection, grants, shared context
```

## Why a host identity is needed

Without host grouping, three Agents on the same daemon can all appear to independently advertise `stt=true` / `tts=true` even though there is only one underlying credential/media provider.

That creates two problems:

### Voice

TTS readiness should make every hosted Agent eligible for independent voice mute/unmute. It should not require duplicate setup or a global one-Agent speaker picker.

### Live Transcript

STT must run only once per Room. If Pi, Codex and Hermes all live on one STT-ready Runtime, the Room must know they represent **one producer host**, not three independent STT producers.

## Minimal Runtime Host identity

Introduce or derive an opaque host identity concept, conceptually:

```text
runtimeHostId = stable opaque id for one local Free4Chat Runtime installation/root
```

Desired properties:

- stable across daemon restart/release upgrade for the same Runtime root where practical;
- not derived from hostname, username, IP, MAC address or other identifying machine metadata;
- safe to expose only as an opaque grouping key inside Room state;
- multiple resident Agent participants on the same daemon/root project the same host id;
- no global account/device registry is introduced.

[...]

Issue #169, opened 2026-08-28
runtime: make resident Agent lifecycle actions truthful and host-owned


madawei2699:
## Status

**COMPLETED — implemented, released, activated, and production-dogfooded.**

The resident lifecycle truthfulness gap is closed in Runtime `0.5.15`.

Required rule now holds:

> **Conversational lifecycle claims must match actual Runtime / Room lifecycle state.**

## Ownership boundary

The implementation preserves the intended separation:

```text
Harness
= intelligence, tools, private memory, local approval policy

Agent Runtime / daemon
= Room participant lifecycle, handle, cursor, lease, reconnect/rejoin,
  media and Harness session ownership

Host / operator environment
= process installation / replacement and ultimate Runtime process lifecycle
```

Useful shorthand:

> **Harness owns intelligence; Runtime owns Room participation; Host owns Runtime process lifecycle.**

No unrestricted process control, remote shell, generic supervisor, background updater, or fleet-management surface was introduced.

## Implementation

Implemented by #217 (`fix(agent): make resident self-leave truthful`) and merged as `25713a97be2464b2644c7403c2a70ed923a5832f`.

The Harness can emit exactly one closed local lifecycle envelope:

```text
[[free4chat:lifecycle leave]]
```

Important properties:

- the envelope is parsed only at the Harness → Runtime boundary;
- only the exact terminal form is recognized;
- ordinary prose, quoted examples, malformed variants, Agent-authored context, transcript content, attachments, and unaddressed turns do not create lifecycle authority;
- `targets` and `lifecycle` controls are mutually exclusive in either ordering and fail closed when combined;
- Runtime additionally requires an explicitly addressed Human event before honoring the leave intent;
- arbitrary Harness success wording is suppressed before authoritative leave confirmation;
- Runtime calls the real `leave_room` path and treats successful Room leave as the authoritative boundary;
- failed leave keeps the resident alive/recoverable and reports fixed truthful failure text instead of a model-supplied success claim;
- confirmed leave closes the rejoin path, clears participation state, and hands blocking cleanup to the daemon asynchronously so `Stop()` cannot self-deadlock on the active wait-loop;
- operator `free4chat-agent leave <instanceId>` behavior remains unchanged.

No Room/MCP protocol change was required.

## Release / activation

Runtime `0.5.15` was published from the #217 merge source.

Release verification completed:

- [x] `agent-v0.5.15` exists;
- [x] darwin-amd64 asset verified;
[...]

Issue #157, opened 2026-08-28
experiment: shared ephemeral Room context across Humans and Agents


madawei2699:
## Status

**EXPERIMENT A PRODUCTION ACCEPTED. Experiment B / #192 is now the current mainline.**

Free4Chat has now proven in production that committed Room-wide Live Transcript is usable as bounded Room-shared ephemeral context by independently running Agents without Human copy/paste, a Notes Agent, or shared STT capability.

The architectural decision from this experiment is now evidence-backed:

> **Bounded Room transcript replay + Harness-side context selection is sufficient for the current shared-context need. Do not add a central Memory/ContextStore/vector database.**

## First principle

> **Room-shared context is collaboration substrate, not participant-owned memory.**

Free4Chat owns bounded availability/transport. Each Harness owns private reasoning, context selection, summarization, cache and durable memory.

Keep these classes separate:

1. **Private participant context** — Harness reasoning/history, credentials, cookies, unshared local files, private tool/model state.
2. **Room-shared ephemeral context** — messages, collab events, progress/decisions, presence/capabilities, artifact references, committed attributed Live Transcript segments.
3. **Explicit artifacts** — screenshots, patches, logs/JSON, files, transcript exports, reviews.

## Experiment A — production result ✅

Production stack:

- #209 merged at `d67c34fbd0a8614000602fb53cdb587903a7d3b0`;
- `agent-v0.5.12` published from that exact source SHA;
- #210 activated the live bootstrap;
- production `agent.md` now installs Runtime 0.5.12.

Validated topology:

```text
Runtime A — independent daemon, STT ready
└─ Agent A

Runtime B — independent daemon, stt=false / tts=false
└─ Agent B

Human
```

Observed production behavior:

- Human speech became committed Room-wide Live Transcript;
- transcript commits did **not** create ordinary Room chat messages;
- transcript commits did **not** wake Agent A or Agent B;
- explicitly targeting Agent A later caused A to consume the committed transcript correctly;
- explicitly targeting Agent B later caused B, on a different Runtime with no STT configuration, to independently consume the same Room transcript;
- B did not depend on A's reply or a Human paraphrase to recover the spoken facts;
- after B left and rejoined, it remained passive until explicitly targeted again;
- after explicit targeting post-rejoin, B recovered the same bounded Room transcript context;
- no private Harness state or credential was required to cross between participants;
[...]

PR #178, opened 2026-08-28
runtime: model Runtime Host speech capabilities (#176 Phase A)

madawei2699:
Part of #176 (Phase A only). **Not merged** — opened for review.

## Why

Multi-Agent dogfooding showed one native Runtime daemon serving several resident Agents while the Room treated speech readiness as if each Agent individually owned it. Three Agents on one daemon could all appear to independently advertise STT/TTS although there is exactly one underlying credential/media provider.

Phase A lands the smallest Runtime Host foundation that #170 (per-participant voice UI/grants), #177 (live transcript producer counting), and #157 (shared context) build on. **Data-only**: no media behavior change, no UI, no transcript, no Human↔Host association yet, and none of the non-goals (device inventory, credential vault, process management, DB/queue/R2/listener infrastructure).

### 1. Private root seed + Room-scoped `runtimeHostId`
- The **private** random seed is generated once and persisted at `<runtimeDir>/host-seed` (0600) under the Runtime root: stable across daemon restarts/release upgrades, shared by every resident of the root, distinct per root.
- Initialization is **atomic and no-clobber** (#178 review fix 1): the complete seed is published with `link(2)` from a unique temp file — no reader can ever observe an empty or partially written file, concurrent first-callers cannot clobber each other. A **50-goroutine race test** proves exactly one seed, complete file, 0600.
- **Never derived** from hostname, username, IP, MAC, or any machine metadata; a malformed leftover file is regenerated (lost grouping key beats a poisoned one).
- The **public** `runtimeHostId` is **derived per Room** as HMAC-SHA256(seed, final non-empty roomId) (#178 review fix 3): same root + Room stable across restarts/rejoins; different Rooms on one root diverge (no cross-Room correlation); different roots diverge; the raw seed is **never exposed**. Create-first derives only after the final server-generated roomId exists (pushed additively after adoption).

### 2. `RuntimeHostProjection {runtimeHostId, speech:{stt,tts}}`
- Coarse, secret-free readiness derived from the local speech config. `stt/tts=true` means "this host can currently produce STT/TTS **if a Room grant authorizes it**". Provider keys, credential sources, and everything else are structurally excluded — the shape has no field for them.
- Discovery metadata only, never authorization; a ready host never starts media by itself.

### 3. Wire path (existing Room primitives only)
- Canonical Room model: `RoomRecord.runtimeHosts` holds **ONE readiness projection per host id**, shared by all same-host Agents; participants reference hosts via `runtimeHostId`.
- `join_room` / `create_room` accept an optional `runtimeHost` object at registration.
- New MCP tool `update_runtime_host` → DO action `agent-update-runtime-host` upserts the shared entry, so speech hot reload (#167) reaches every same-host Agent **without any resident rejoining**.
- **Additive** (#178 review fix 5): Agent-only; malformed registration payloads are dropped and the text join proceeds (Humans and Agents alike); update-op failures fail loudly without corrupting the last valid projection; update-push failures log bounded reason codes only. Projections run through `room-info`, browser state, and the `wait_for_events` roster response (Go `WaitResult` parses the shared map strictly); host projections are garbage-collected when their last Agent departs; malformed persisted entries are dropped and dangling references cleared on load. Presence/discovery broadcast only — no messages, no sequence churn, no waiter wakes.
- Every (re)join re-projects the Runtime's **own** identity — reconnect/rejoin can never inherit another host's state. Humans never project a host.

### 4. Docs
- `app/public/agent.md` canonical tool list (15→16 tools) and the MCP developer page document the new tool and the optional registration parameter.

## Tests

Go (`hostid_test.go`, `client_test.go`, `runtime_test.go`, `daemon_test.go`):
1. seed stable within a root; file 0600; distinct roots → distinct seeds; malformed file regenerated; **50-goroutine concurrent first-use yields exactly one seed** (atomic no-clobber).
2. `DeriveRuntimeHostID`: deterministic per (root, Room); never equals the raw seed; different Rooms/roots diverge; fails closed without a final roomId.
3. `join_room`/`create_room` carry the derived projection only when present and valid — legacy payload byte-identical; `update_runtime_host` payload exact.
4. Runtime: join carries the derived id; speech hot reload pushes updated readiness without rejoining; seedless runtimes join without any projection (text never blocked).
5. Daemon `status` exposes the derived `runtimeHostId` + `speech` per resident (never the raw seed).
6. `ParseRuntimeHostStrict` fails closed: 12 negative cases (missing/invalid speech, bad id charset/length, non-object, junk envelope).

DO (`roomSessionRuntimeHost.test.ts`, 6):
7. register with projection → canonical map + participant id; room-info, wait roster, and shared map consistent; Humans never project.
8. **two same-host Agents share ONE readiness projection**; hot reload via one Agent updates the shared entry for both.
9. malformed registration payloads dropped and the join proceeds (additive); no-projection Agents unaffected; Humans never get a host id.
10. `agent-update-runtime-host`: valid upsert; wrong token 401; human 403; malformed 400 with last valid projection surviving.
11. host projections garbage-collected when their last Agent departs; shared hosts kept while referenced.
12. persistence across loads; malformed map entries dropped and dangling participant references cleared (storage repaired).

## Validation

- Go: `gofmt -l` clean, `go vet ./...` clean, `go test ./... -count=1` 11/11 packages, `go build ./cmd/free4chat-agent`, linux + darwin (CGO) cross-builds, `scripts/check-cleanup.sh` OK.
- App: `prettier --check`, `eslint --max-warnings 0`, `tsc --noEmit`, `vitest` 37 files / 412 tests, `next build`.

## Out of scope (later phases)

#170 per-participant voice UI/grants · #177 live transcript · Human↔Runtime association mechanism · #169 lifecycle ownership · any UI change.


madawei2699:
All five review blockers fixed in `6749073` (single commit, 18 files). No scope change: #170/#177, ownership proof, voice/media, version/release untouched; #175 behavior/tests preserved.

**1. Atomic, no-clobber seed init.** Seed content is written fully to a unique temp file and published with `link(2)` (fails EEXIST — no clobber, no empty-file window); loser reads the winner's seed. `TestRuntimeHostSeedConcurrentInitializationYieldsOneSeed` races **50 goroutines** on one fresh root and proves exactly one seed, complete file, 0600.

**2. Strict parser.** `ParseRuntimeHostStrict` enforces the wire contract fail-closed: id `/^[A-Za-z0-9._:-]{8,64}$/`, required speech object, required stt/tts booleans; malformed/custom/stale payloads → nil. `TestParseRuntimeHostStrictFailsClosed` pins 12 negative cases (missing speech, non-object speech, missing/nil/non-bool slots, id space/short/long/empty/non-string, non-object, junk envelope).

**3. Room-scoped public identity.** The raw root UUID is now a PRIVATE `host-seed` (0600) that never leaves the machine or enters Room state; the public `runtimeHostId` = HMAC-SHA256(seed, final non-empty roomId) — derived per Room. Same root+Room stable across restarts/rejoins; different Rooms/roots diverge; create-first derives only AFTER the server-generated roomId exists (pushed additively post-adoption). `TestDeriveRuntimeHostIDRoomScoped` covers same/different room + root; status views show only the derived id.

**4. Canonical Room model.** `RoomRecord.runtimeHosts` holds ONE readiness projection per host id; participants reference hosts via `runtimeHostId`. room-info, browser state, and the wait_for_events roster response project the shared map. Updates upsert the shared entry — hot reload reaches every same-host Agent; host projections are GC'd when their last Agent departs; loads sanitize malformed map entries and clear dangling references. Tests: two same-host Agents sharing one projection, hot reload for both, persistence, leave GC.

**5. Additive failures.** Malformed registration payloads are dropped and the text join proceeds (Humans and Agents); seed-load failure joins without a projection; update-push failures log bounded reason codes only (no seed/handle/credential material); the client omits an invalid projection instead of sending it. `TestRuntimeWithoutSeedJoinsWithoutHostProjection` pins the legacy path.

**Validation:** gofmt/vet clean, `go test ./... -count=1` 11/11 packages, `go build` + linux/darwin(CGO) cross-builds, cleanup contract OK; prettier/eslint/`tsc`/`vitest` 37 files 412 tests/`next build` all green. Author/committer `codex <267193182+codex@users.noreply.github.com>`.

Issue #317, opened 2026-09-10
perf(cost): remove resident control polling and Room DO write amplification


madawei2699:
## Status

**PR A SHIPPED AND PRODUCTION-VERIFIED. PR B IS DEFERRED UNTIL RESIDUAL COST EVIDENCE JUSTIFIES IT.**

#239 / PR #240 previously moved official resident delivery from HTTP long-polling to the hibernatable Agent event WebSocket.

This issue then identified a second dominant cost source: old resident Runtime media Controllers polling `room_info` every 5 seconds through `POST /room/control`.

PR #319 removed that recurring polling path for official residents and shipped in Agent Runtime 0.5.27.

## Production result — 0.5.27

The source of the abnormal fixed-rate `/room/control` traffic is now confirmed.

Observed before upgrading/stopping old residents:

~~~text
multiple resident Agents
→ approximately every 5 seconds
→ one fixed group of /room/control requests
~~~

After stopping the old Runtime residents:

~~~text
wait + refresh Cloudflare realtime logs
→ fixed 5-second request pattern disappeared
~~~

Then a real 0.5.27 resident was joined to the production `test` Room:

~~~text
Runtime 0.5.27 joins Room
→ browser sees Runtime-0527-Test
→ directed Human message
→ Agent replies RESIDENT-0527-OK
→ resident event loop remains waiting/healthy
→ no recurrence of the old 5-second /room/control pattern during the observation window
~~~

The test resident was then stopped to avoid creating unnecessary production usage.

This establishes the primary acceptance condition:

> **An idle/current official resident no longer performs recurring 5-second `room_info` polling.**

The previous high request/write metrics were polluted by long-running old Runtime residents, so they should not be used to size follow-up optimization without a clean 0.5.27 observation window.

## Track A — complete

Implemented by #319:

~~~text
existing resident event WebSocket
→ bounded self-targeted media-state projection
→ Runtime media Controller reconciliation
→ no recurring 5-second RoomInfo ticker
~~~

Verified properties include:

- initial pushed media projection;
- grant/revocation reconciliation;
- reconnect requires fresh state;
- fail-closed media behavior on transport/projection loss;
- media-only state does not wake Harness cognition;
- ordinary resident event-loop / directed work still functions.

## Remaining possible work — evidence gated

The original investigation also identified possible write amplification in:

- resident heartbeat/liveness persistence;
- unconditional Room + Live Transcript persistence;
- redundant alarm scheduling;
[...]

PR #319, opened 2026-09-10
[codex] perf(cost): remove resident media polling

madawei2699:
## Summary

- Refs #317.
- Remove the resident Agent Runtime's recurring `room_info` polling for media authorization.
- Reuse the existing hibernatable resident event WebSocket to deliver a bounded, self-targeted media projection.

## Root cause / cost path

The resident media Controller previously polled `room_info` every 5 seconds. Each poll is a `POST /room/control` request that wakes the Room Durable Object and performs a storage-backed observation. With three resident runtimes, this accounts for roughly 2,100 DO control requests per hour before normal room traffic. The separate SFU track-discovery cadence is unchanged by this PR.

## What changed

- Add a minimal resident-only `mediaState` projection to the existing Agent event stream.
- Push the current state on initial connection, reconnect, grant transitions, and environment media-gate changes.
- Reconcile Meeting Notes, Agent Voice, and Live Transcript state from the pushed projection.
- Fail closed when the authenticated event stream fails or an envelope lacks the required projection; require a fresh state after reconnect before media can run.
- Keep media-only envelopes outside the Runtime event queue so they never wake the Harness.
- Retain one-shot `RoomInfo` bootstrap compatibility for non-resident/injected clients, but remove the recurring Controller ticker.

## Invariants

- The projection is self-targeted and does not expose another Agent's grant, Human identifiers, handles, tokens, or media/session identifiers.
- The public `wait_for_events` contract is unchanged.
- Existing SFU subscription polling remains separate and unchanged.
- PR B (heartbeat/persistence write amplification) is intentionally not included.

## Verification

- `yarn test` — 75 files, 671 tests passed.
- `yarn type-check` passed.
- `yarn docs:check` passed.
- `yarn eslint src --no-fix` passed.
- `yarn prettier --check .` passed.
- `yarn cf-build` completed successfully; OpenNext emitted existing non-fatal copy warnings.
- `go test ./...` passed.
- `go vet ./...` and `go build ./cmd/free4chat` passed.
- Focused resident event-stream race test passed.
- CI-equivalent `go test ./... -count=1` passed after the follow-up fix.

## Follow-up fix

- Start pushed-state Controllers synchronously before opening the resident event stream, closing the initial projection scheduling gap.
- Retain the last authenticated projection only in Runtime memory for live Controller rebuilds such as speech hot reload; clear it on transport loss before reconnect.
- Align the Live Transcript compatibility test fixture with the production `AGENT_MEDIA_ENABLED` projection contract.

After deployment, production DO request/write metrics should be rechecked to confirm the expected reduction and to size the separate PR B follow-up.

### DIFF
--- a/agent/internal/free4chat/resident_events.go
+++ b/agent/internal/free4chat/resident_events.go
@@ -36,6 +36,7 @@ type residentEventEnvelope struct {
 	ExpiresAt    int64                      `json:"expiresAt"`
 	Participants []json.RawMessage          `json:"participants"`
 	RuntimeHosts map[string]json.RawMessage `json:"runtimeHosts"`
+	MediaState   *types.ResidentMediaState  `json:"mediaState,omitempty"`
 	Expired      bool                       `json:"expired,omitempty"`
 	Truncated    bool                       `json:"truncated,omitempty"`
 }
@@ -135,9 +136,10 @@ func (s *residentEventStream) Receive(ctx context.Context) (types.WaitResult, er
 		return types.WaitResult{}, &Error{Message: "resident event stream returned an invalid envelope", Code: CodeToolError}
 	}
 	wait := types.WaitResult{
-		Events:    envelope.Events,
-		Cursor:    envelope.Cursor,
-		ExpiresAt: envelope.ExpiresAt,
+		Events:     envelope.Events,
+		Cursor:     envelope.Cursor,
+		ExpiresAt:  envelope.ExpiresAt,
+		MediaState: envelope.MediaState,
 	}
 	if envelope.Participants != nil {
 		raw := make([]any, 0, len(envelope.Participants))
--- a/agent/internal/free4chat/resident_events_test.go
+++ b/agent/internal/free4chat/resident_events_test.go
@@ -58,6 +58,14 @@ func TestResidentEventStreamUsesHeadersAndDecodesEnvelope(t *testing.T) {
 					"speech":        map[string]any{"stt": true, "tts": false},
 				},
 			},
+			"mediaState": map[string]any{
+				"meetingNotes":        map[string]any{"active": true, "startedAt": 11},
+				"agentVoiceEnabledAt": float64(22),
+				"mediaAvailable":      true,
+				"liveTranscript": map[string]any{
+					"active": true, "producerRuntimeHostId": "host-a", "epoch": float64(7),
+				},
+			},
 		})
 		_ = conn.Write(context.Background(), websocket.MessageText, payload)
 	}))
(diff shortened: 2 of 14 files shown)


madawei2699 (commented):
Review of head `fbfc195a7edf109f265ecb4d64b3b355d5261e3b`: **CHANGES REQUESTED**.

PR A direction is correct and the architecture boundary is good: the 5s `room_info` ticker is removed for official residents, the new media projection is a narrow pure module, `RoomSession` only gains the projection at the existing resident-envelope boundary, and `useSfuChatRoom` / `RoomContent` are untouched.

Two blockers remain:

1. **Initial pushed media state can be lost in a startup race.** `NewController()` starts with `stopped=true`. `restartMediaController()` publishes `r.mediaController = controller` and then starts it asynchronously with `go controller.Start(...)`. Once `StartLoop()` opens the resident event stream, an initial `mediaState` envelope can reach `ObserveMediaState()` before that goroutine runs; `observeMediaState()` returns immediately while `stopped`, so the authorization snapshot is discarded. `Start()` then sets `stopped=false` and, because `UsePushedMediaState=true`, does no bootstrap. With no later Room state change, an already-active Meeting Notes / Agent Voice / Live Transcript grant can remain locally inactive indefinitely. Make the pushed-state controller ready synchronously before the resident stream can deliver its first envelope (or buffer/replay a pre-Start observation), and add a deterministic regression test.

2. **Current head has a failing required Go CI check.** `Agent Runtime (Go)` run `34429846587` fails `go test ./... -count=1` in `TestControllerLiveTranscriptElectsOneVerifiedSameHostProducer` (`timeout waiting for first live producer`). The changed `residentMediaStateFromRoomInfo()` gates Live Transcript through `MediaAvailable = MeetingNotesMediaAvailable || AgentVoiceMediaAvailable`, while the previous compatibility path did not apply that combined gate to Live Transcript. Production `room_info` currently projects both availability booleans from the same `AGENT_MEDIA_ENABLED` gate, so either make the test fixture reflect that server contract or preserve the intended compatibility semantics explicitly; do not leave the required Go workflow red.

PR B should remain separate; no request to add heartbeat/persistence work here.

madawei2699 (commented):
Review of latest head `4b911e4e5ec65c84984b8c71d6761e18da0cc996`: **CHANGES REQUESTED — one new blocker.**

The two prior blockers are fixed:

- startup ordering is now correct for the initial resident projection: pushed-state Controller `Start()` is synchronous before `StartLoop()` can open the event stream;
- the Live Transcript compatibility fixture now matches the production `AGENT_MEDIA_ENABLED` availability contract, and all four current workflows are green, including `Agent Runtime (Go)`.

The PR still has one authorization-ordering race introduced by the new hot-reload replay cache:

### Stale cached media projection can be replayed after fail-closed / a newer projection

`replayResidentMediaState()` snapshots `{controller, state, valid}` under locks, releases the locks, and only then calls `controller.ObserveMediaState(state)`.

`failClosedResidentMediaState()` and `observeResidentMediaState()` use the same pattern: update/snapshot under locks, then call the Controller after releasing them.

This permits an ordering such as:

```text
ReloadSpeech → restart Controller C2
replayResidentMediaState snapshots old ACTIVE state and valid=true
                                    [pause]
resident event stream disconnects
→ failClosedResidentMediaState clears cached state
→ C2.FailClosedMediaState() stops media
                                    [resume]
replay → C2.ObserveMediaState(old ACTIVE)
→ stale authorization is re-enabled while the authenticated stream is down
```

The same race exists against a newer pushed revocation/update: a stale replay can land after the newer state and revert the Controller.

This directly violates the PR's fail-closed invariant, so it is a blocker rather than a test-only concern.

Please serialize Runtime-level resident media-state **application ordering** across observe / fail-closed / replay (or use an equivalent generation/revision scheme that makes stale replay impossible). Avoid holding `residentMediaStateMu` across Controller reconciliation if that creates lock-order inversion with `mediaMu` callbacks. A small dedicated application-order mutex is one possible narrow shape, but use the simplest safe implementation after checking existing lock ordering.

Add deterministic regressions for at least:

1. cached ACTIVE replay racing transport fail-closed → final Controller remains inactive;
2. cached old state replay racing a newer pushed revocation/state → newest observation wins.

Everything else in PR A remains appropriately scoped. Do not pull PR B heartbeat/persistence work into this fix.

Issue #314, opened 2026-09-10
task: isolate Agent context and allow explicit Agent participation


madawei2699:
## Status

**COMPLETED — implemented by #320, released in Agent Runtime 0.5.27, and production-dogfooded.**

This issue closed the remaining Task cognition-boundary gap found after the first task-scoped interaction rollout.

## Shipped behavior

The production model is now:

~~~text
Task T primary Agent X
→ X receives T as scope=task:T

unrelated Agent Z
→ does not receive T as ordinary Room context

Human in Task T explicitly targets Agent Y
→ Y receives addressed scope=task:T
→ Runtime creates/reuses Y:T
→ X:T and Y:T remain independent Harness conversations
~~~

The implementation preserves the intended asymmetry:

~~~text
Human Room participant
→ may open/read/interact with Room-visible Tasks

Agent
→ must participate explicitly in Task cognition
→ unrelated Task content never degrades into Room context
~~~

No Task Human membership, unread model, notification system, Thread entity, or Task membership database was added.

## Lifecycle truth

Task lifecycle remains canonical rather than inferred from prose:

~~~text
request   → Starting
accepted  → Working
completed → Completed
failed    → Failed
~~~

Runtime 0.5.27 production dogfood observed the canonical `accepted` transition and `Status Working` before the Agent completed work.

## Production evidence

0.5.27 dogfood verified:

- Codex-targeted Task work stayed targeted;
- explicit Task `@Pi` successfully created/used Pi's Task scope and returned the expected response;
- Human Task `accepted → Working` worked in the real Room;
- ordinary resident Room turns remained functional;
- task-scope isolation and reconstruction paths remain covered by automated tests.

The private Harness context of an unrelated Agent is intentionally not exposed to the browser merely for observability; the fail-closed task projection behavior is covered in code/tests instead of adding a new introspection surface.

## Architecture result

Task visibility/participation decisions were kept behind the narrow Task projection seam rather than growing a new Room-level membership/state machine.

The canonical logical session model remains:

~~~text
(roomId, agentParticipantId, scopeId)

scopeId = room | task:<requestId>
~~~

## Follow-up

Per-scope execution concurrency is intentionally separate and evidence-gated in #315.
Agent activity UX is tracked in #324.
Task Live View is tracked in #316.

Related: #300, #309/#312, #310/#311, #315, #324, #316.

PR #328, opened 2026-09-10
feat: show transient Agent activity

madawei2699:
## Summary

- project coarse transient Agent activity (`Working`, `Thinking`, `Using tools`, `Responding`) from the canonical Go ACP Runtime into the Room UI
- keep activity scoped to the authenticated Agent and `room` / owned `task:<requestId>` scope
- add the narrow authenticated Runtime-to-Room activity control route, including the Worker route allowlist entry
- keep activity in transient DO memory only; it does not enter the Room message log, resident event stream, Harness wakeup path, or analytics
- clear activity on turn completion/failure, Harness failure, Runtime stop, resident stream loss, reconnect, Agent departure, and lease expiry
- preserve serialized Harness turns, existing media event-loop behavior, and the PR #317 no-polling design

## Validation

- `cd app && yarn test` — 78 files, 686 tests passed
- `cd app && yarn type-check` — passed
- `cd app && yarn eslint src --no-fix` — passed
- `cd app && yarn prettier --check .` — passed
- `cd app && yarn docs:check` — passed
- `cd app && yarn cf-build` — passed
- `cd agent && go test ./... -count=1` — passed
- `cd agent && go vet ./...` — passed
- `cd agent && go build ./...` — passed

### DIFF
--- a/agent/internal/runtime/activity.go
+++ b/agent/internal/runtime/activity.go
@@ -34,13 +34,61 @@ func validActivityScope(scope string) bool {
 	return strings.HasPrefix(scope, "task:") && len(strings.TrimPrefix(scope, "task:")) >= 4
 }
 
+type activityPublication struct {
+	handle string
+	state  types.AgentActivityState
+}
+
+// publishActivity enqueues only the newest state for each scope. It is called
+// from the ACP reader and serialized turn path, so it must never wait for
+// Room/HTTP I/O.
 func (r *ResidentRuntime) publishActivity(scope string, state types.AgentActivityState) {
-	client, ok := r.options.Client.(types.ResidentActivityClient)
-	if !ok {
-		return
-	}
 	handle := r.currentHandle()
 	if handle == "" {
+		return
+	}
+	r.activityPublishMu.Lock()
+	if r.activityPublishQueue == nil {
+		r.activityPublishQueue = make(map[string]activityPublication)
+	}
+	r.activityPublishQueue[scope] = activityPublication{handle: handle, state: state}
+	if r.activityPublisherActive {
+		r.activityPublishMu.Unlock()
+		return
+	}
+	r.activityPublisherActive = true
+	r.activityPublishMu.Unlock()
+	go r.drainActivityPublications()
+}
+
+// drainActivityPublications is deliberately a small, per-Runtime publisher,
+// not a general event bus. A single sender preserves application order for a
+// scope: if an old state is already in flight, a later clear waits behind it;
+// if it is still queued, latest-state replacement removes it. Thus a final
+// clear cannot be overtaken by stale Working/Thinking network traffic.
+func (r *ResidentRuntime) drainActivityPublications() {
+	for {
+		r.activityPublishMu.Lock()
+		if len(r.activityPublishQueue) == 0 {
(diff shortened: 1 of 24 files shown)


madawei2699 (commented):
Review conclusion for head d8e83b7d026e226ce9ec099686e95d034a745fbe: CHANGES REQUESTED before merge.

Three concrete blockers/fixes remain, matching the inline review threads:

1. **P1: activity HTTP must not block Harness/ACP progress.** `publishActivity()` is synchronous from `beginActivity`/`finishActivity`, and rich ACP updates reach it from the sole ACP stdout dispatch path. Slow/unavailable `/api/room/agent-activity` can therefore delay the turn and stop ACP responses/permission traffic from being consumed. Activity is best-effort presentation; move publication behind a bounded non-blocking/coalescing path.

2. **Task secondary-Agent activity is hidden.** Server-side canonical task ownership correctly accepts an explicitly participating Agent, but `RoomContent` only permits the original task creator/target. After #314, `Task T @Pi` must show Pi:T activity. Render activity for any connected Agent whose already-authorized projection belongs to the active task, including multiple Agents when applicable.

3. **Clear the transient map on Room generation expiry/reset.** Otherwise stale entries survive a same-name Room recreation on a still-live DO and keep counting toward `MAX_AGENT_ACTIVITY_ENTRIES`.

Keep the fixes narrow: no polling, no durable activity persistence, no new event bus, no Runtime concurrency, no #317 PR B. Add focused regressions for non-blocking activity transport, secondary/multiple Task Agent rendering, and Room-expiry cleanup.

agent/internal/runtime/activity.go
```diff
@@ -0,0 +1,186 @@
...
+}
+
+func (r *ResidentRuntime) resetActivityLocal() {
+	r.activityMu.Lock()
+	r.activities = make(map[string]types.AgentActivityState)
+	r.activityTurnActive = false
+	r.activityScope = ""
+	r.activityMu.Unlock()
+	r.activityPublishMu.Lock()
+	r.activityPublishQueue = nil
```
madawei2699: **P1 — preserve the queued clear across resident reconnect.** `residentWaitLoop()` calls `clearActivity()` after a stream loss, which queues a final `activity=null`; then a successful reconnect calls `setResidentStream()` → `resetActivityLocal()`, and this line drops that queued clear. If an older Working/Thinking HTTP request is already in flight, it can complete *after* the server-side reconnect cleared `transientAgentActivities`. The resident event reconnect intentionally reuses the same participant token while the lease is valid, and `/agent-activity` authenticates that token rather than the WebSocket connection nonce, so the delayed old request is still accepted and can resurrect stale activity with no later clear. Binding the publication to the old handle does not fence this case because the handle remains valid across normal stream reconnects. Preserve ordering so the clear remains behind any in-flight old state (or add an equivalent connection-generation fence), and add a regression that blocks the old publication, reconnects, releases it, then verifies the final public activity is empty.

madawei2699 (commented):
Second-pass review of head `edd250799b485af5d332391a73752c4ad495733a`: the original three fixes are largely correct, but one reconnect race remains in the async publisher. Please fix the stale in-flight publication case described inline; after that I expect this PR to be merge-ready.

Issue #344, opened 2026-09-11
experiment: sandboxed Room Apps over the existing realtime Room substrate


madawei2699:
## Status

**PHASE 0 STATIC ARCHITECTURE PROOF PASSED — production deployment/dogfood is active under #361. Keep this umbrella open until the Phase 0 decision is recorded.**

Room Apps test one implementation hypothesis:

> **Large realtime collaborative experiences can enter a temporary Room as sandboxed external Apps while Free4Chat remains a small rendezvous / participant / transport core.**

Phase 0A has now produced positive evidence:

- #356 merged the generic sandbox host + bounded reliable/realtime bridge;
- #354 closed;
- `free4chat-extension-lab#14` Shared Canvas merged;
- `free4chat-extension-lab#15` Tiny Arena merged;
- both fixtures use the same seam;
- neither required Canvas/Arena domain logic in Free4Chat core;
- no second realtime backend or per-App PeerConnection was introduced.

This proves the seam is technically plausible across two materially different applications. It does **not** yet prove production runtime quality, cost shape or product retention value.

## Current execution

#361 owns the remaining Phase 0B work:

```text
deploy merged Lab fixtures to the curated production origin
→ independently smoke-test both HTTPS paths
→ enable the existing ROOM_APPS_ENABLED kill switch
→ real 2–4 Human desktop/mobile Room dogfood
→ validate late join / leave / close-reopen / chat+voice coexistence
→ measure coarse realtime traffic / SFU cost direction
→ return here for PROCEED / REVISE / STOP
```

The curated paths are already fixed by the host contract:

```text
https://room-apps.free4.chat/shared-canvas
https://room-apps.free4.chat/tiny-arena
```

Do not redesign the bridge merely to complete deployment.

## Stable architecture boundary

```text
Free4Chat core
├─ Room lifecycle / participant identity + presence
├─ curated App identity/origin
├─ sandboxed iframe host
├─ app instance lifecycle
├─ bounded participant projection
├─ logical reliable + realtime lanes
└─ payload/rate/cost guardrails + kill switch

Room App
├─ rendering
├─ domain rules
├─ state / CRDT / simulation
├─ interaction semantics
├─ optional external backend/persistence
└─ optional semantic Agent interface
```

Free4Chat core must not learn App concepts such as `stroke`, `shape`, `entity`, `position`, `rotation`, `HP`, `score`, board state or game economy.

## Task Live View != Room App

Keep the execution models separate:

```text
Task Live View
[...]

PR #356, opened 2026-09-11
Add Phase 0 Room App host and bounded realtime bridge

### LINKED ISSUE #354: room-app phase 0a: experimental sandbox host and bounded realtime bridge
## Status

**READY — implementation prerequisite for the two parallel Phase 0 fixtures in #344.**

Parent direction: #344

Lab fixtures:
- `i365dev/free4chat-extension-lab#12` — Shared Canvas / reliable lane
- `i365dev/free4chat-extension-lab#13` — Tiny Three.js Arena / realtime lane

This issue is deliberately narrower than a Room App framework. It exists only to provide one experimental seam that both fixtures can consume.

## Goal

Prove the smallest Free4Chat-side host/transport boundary needed to mount a curated sandboxed external App and exchange bounded logical realtime messages over the existing Room/SFU/DataChannel substrate.

After this seam is usable, the Shared Canvas and Tiny Arena fixture implementations may proceed in parallel.

## Experimental boundary

```text
Free4Chat Room
→ selected App interaction
→ sandboxed allowlisted iframe
→ postMessage bootstrap/handshake
→ dedicated MessagePort
→ bounded participant projection
→ logical reliable lane
→ logical realtime lane
```

This is an **internal experimental contract**, not a public SDK.

## Minimal host responsibilities

Free4Chat may own only:

- curated/allowlisted App identity/origin;
- App instance lifecycle;
- iframe sandbox host;
- bootstrap/handshake;
- dedicated MessagePort;
- bounded `self` and participant projection;
- logical reliable/realtime routing keyed by App instance;
- payload/rate/instance guardrails;
- coarse transport counters needed for cost verification;
- kill switch / feature flag.

Free4Chat must not own application-domain state or rules.

## Credential/security boundary

The iframe must not receive:

- Room participant token;
- SFU credentials;
- raw PeerConnection/DataChannel objects;
- provider/Agent credentials;
- arbitrary Room history;
- Free4Chat cookies/auth state;
- filesystem/microphone/camera capabilities by default.

The parent bridge is the trust boundary.

## Bridge shape

Names remain provisional. Keep the implementation easy to replace.

Conceptually:

```text
host → app
ready/self/participants
participant_join
participant_leave
reliable(payload)
realtime(payload)

app → host
ready
sendReliable(payload)
sendRealtime(payload)
```

Every wire message must be bounded and associated with the current App instance. Do not accept arbitrary relay destinations.

## Transport rules

- Reuse the existing participant realtime substrate.
- Do not create one new PeerConnection per App by default.
- Do not route high-frequency App frames through Durable Object Room messages/history/storage.
- Reliable lane is for low-frequency state transitions/ops that must arrive.
- Realtime lane is for stale-droppable transient updates and should prefer newer state.
- Do not expose raw transport internals to the iframe.

## Guardrails

Before fixture dogfood, define conservative experimental caps for:

- payload bytes;
- messages/sec;
- bytes/sec;
- active App instances per Room/participant;
- malformed/unknown App instance messages.

Add a simple feature flag / kill switch so the experiment can be disabled without touching ordinary Room collaboration.

Do not over-optimize or build a generic quota framework.

## UI scope

Reuse the current Room interaction/Stage seam. Add the smallest App interaction mounting path necessary.

Do not redesign navigation or turn `RoomContent` into a generic plugin framework.

A second fixture must be able to use the same seam without adding a second domain-specific branch.

## Testing

Add focused tests for:

- allowlisted App origin only;
- sandbox attributes/capability boundary;
- valid handshake and MessagePort ownership;
- rejecting malformed/wrong-app-instance messages;
- participant projection contains only intended bounded fields;
- reliable vs realtime routing stays logically separated;
- payload/rate caps;
- App unmount/close cleanup;
- Room remains usable if App handshake fails or iframe disappears;
- ordinary Room voice/chat/tasks/screenshare regressions remain green.

Do not add Canvas/Three.js domain tests here.

## Acceptance

- [ ] one curated sandboxed iframe can mount in a Room interaction;
- [ ] iframe receives no Room/SFU/provider credentials;
- [ ] iframe gets a bounded self/participant projection;
- [ ] dedicated MessagePort handshake works;
- [ ] logical reliable lane works over existing realtime substrate;
- [ ] logical realtime lane works over existing realtime substrate;
- [ ] no new PeerConnection per App by default;
- [ ] no high-frequency DO/history/storage route;
- [ ] conservative traffic guardrails + kill switch exist;
- [ ] App unmount/failure does not destabilize ordinary Room collaboration;
- [ ] both extension-lab #12 and #13 can branch from this same seam without core changes specific to either domain.

## Non-goals

- stable JS SDK;
- public App manifest;
- registry/marketplace;
- arbitrary URL loading;
- persistent App state;
- App backend hosting;
- product analytics beyond later bounded lifecycle/meaningful-use signals under #346;
- Whiteboard product implementation;
- game product implementation;
- Agent semantic actions.

## Parallel execution

Recommended flow:

```text
this issue / Phase 0A
→ merge/freeze one experimental seam
→ parallel worktrees:
   ├─ extension-lab#12 Shared Canvas
   └─ extension-lab#13 Tiny Arena
→ integrate/dogfood both
→ review complexity + cost against #344
```

If either fixture needs the host contract changed after the split, make that change generically and keep both fixtures green before merging it.

## Kill condition

Stop/revise if supporting the fixtures requires App-specific concepts or switches throughout `RoomSession`, `useSfuChatRoom`, `RoomContent`, a new application server, a new realtime backend, or a new PeerConnection per App.

madawei2699:
## Summary

Refs #344
Closes #354

Adds the narrow Phase 0A experimental Room App seam for the Shared Canvas and Tiny Arena fixtures:

- curated, exact allowlisted App catalog with a sandbox="allow-scripts" iframe;
- dedicated MessagePort bootstrap with one-time handshake token and bounded `self`/`participants` projection;
- generic `reliable` and `realtime` logical lanes multiplexed by `appInstanceId`;
- reuses the existing Human PeerConnection/SFU DataChannel transport: ordered reliable lane and unordered `maxRetransmits: 0` realtime lane;
- simple UTF-8 payload/rate/instance caps, kill switch, in-memory coarse counters, participant join/leave projection, and cleanup/failure isolation;
- short internal Lab-facing contract at `docs/room-app-phase0-contract.md`.

No Room App traffic enters Room messages, Durable Object storage, alarms, analytics, or a per-App PeerConnection. No credentials, tokens, SFU objects, Room history, or Agent state cross into the iframe.

## Validation

- `yarn test`: 83 files / 742 tests passed
- `yarn type-check`: passed
- `yarn eslint src --no-fix`: passed
- `yarn prettier --check .`: passed
- `yarn docs:check`: passed
- `git diff --check`: passed
- focused Room App host/common/hook tests: passed
- `yarn cf-build`: OpenNext reaches the existing repository baseline strict-null type error in `src/common/utils.tsx:34`; no Room App build error was observed.

This PR is intentionally host/bridge only. Canvas and Arena domain implementations remain in extension-lab and must consume this same seam after review. It is not merged.

### DIFF
--- a/app/src/common/roomApp.test.ts
+++ b/app/src/common/roomApp.test.ts
@@ -6,9 +6,11 @@ import {
   decodeRoomAppClientMessage,
   decodeRoomAppEnvelope,
   encodeRoomAppEnvelope,
+  isRoomAppInstanceForRoom,
   isRoomAppAllowlisted,
   projectRoomAppParticipants,
   roomAppRateGuard,
+  roomAppInstanceId,
   validateRoomAppDefinition,
 } from "./roomApp"
 
@@ -51,6 +53,13 @@ describe("Room App Phase 0 bridge contract", () => {
         payload: { text: "界".repeat(ROOM_APP_MAX_PAYLOAD_BYTES) },
       })
     ).toBeNull()
+    expect(
+      encodeRoomAppEnvelope({
+        appInstanceId: "shared-canvas:abc123",
+        lane: "reliable",
+        payload: { sourceParticipantId: "spoofed" },
+      })
+    ).toBeNull()
     expect(decodeRoomAppEnvelope(JSON.stringify({ type: "bogus" }))).toBeNull()
   })
 
@@ -77,6 +86,18 @@ describe("Room App Phase 0 bridge contract", () => {
     ).toBeNull()
   })
 
+  it("accepts only curated instances for the current Room", () => {
+    expect(isRoomAppInstanceForRoom("room-a", "shared-canvas:00000000")).toBe(
+      false
+    )
+    expect(
+      isRoomAppInstanceForRoom(
+        "room-a",
+        roomAppInstanceId("room-a", "shared-canvas")
+      )
+    ).toBe(true)
+  })
+
   it("bounds the participant projection and separates reliable/realtime rate", () => {
     expect(
       projectRoomAppParticipants([
@@ -93,5 +114,11 @@ describe("Room App Phase 0 bridge contract", () => {
     expect(guard.allow("reliable", 1, 1000)).toBe(false)
     expect(guard.allow("realtime", 1, 1000)).toBe(true)
     expect(guard.allow("reliable", 1, 2001)).toBe(true)
+  })
+
+  it("prunes stale bytes from both lanes before applying the shared budget", () => {
+    const guard = roomAppRateGuard()
+    expect(guard.allow("realtime", 200_000, 1000)).toBe(true)
+    expect(guard.allow("reliable", 100_000, 2001)).toBe(true)
   })
 })
--- a/app/src/common/roomApp.test.ts
+++ b/app/src/common/roomApp.test.ts
@@ -1,0 +1,97 @@
+import { describe, expect, it } from "vitest"
+
+import {
+  ROOM_APP_CATALOG,
+  ROOM_APP_MAX_PAYLOAD_BYTES,
+  decodeRoomAppClientMessage,
+  decodeRoomAppEnvelope,
+  encodeRoomAppEnvelope,
+  isRoomAppAllowlisted,
+  projectRoomAppParticipants,
+  roomAppRateGuard,
+  validateRoomAppDefinition,
+} from "./roomApp"
+
+describe("Room App Phase 0 bridge contract", () => {
+  it("accepts only curated app definitions and rejects arbitrary origins", () => {
+    expect(
+      validateRoomAppDefinition({
+        id: "shared-canvas",
+        label: "Shared Canvas",
+        url: "https://room-apps.free4.chat/shared-canvas",
+        origin: "https://room-apps.free4.chat",
+      })
+    ).toBe(true)
+    expect(isRoomAppAllowlisted(ROOM_APP_CATALOG[0])).toBe(true)
+    expect(
+      isRoomAppAllowlisted({
+        id: "user-app",
+        label: "User app",
+        url: "https://evil.example/app",
+        origin: "https://room-apps.free4.chat",
+      })
+    ).toBe(false)
+  })
+
+  it("keeps transport envelopes bounded and UTF-8 sized", () => {
+    const encoded = encodeRoomAppEnvelope({
+      appInstanceId: "shared-canvas:abc123",
+      lane: "reliable",
+      payload: { type: "stroke", points: [[1, 2]] },
+    })
+    expect(encoded).toBeTruthy()
+    expect(decodeRoomAppEnvelope(encoded)).toMatchObject({
+      lane: "reliable",
+      appInstanceId: "shared-canvas:abc123",
+    })
+    expect(
+      encodeRoomAppEnvelope({
+        appInstanceId: "shared-canvas:abc123",
+        lane: "reliable",
+        payload: { text: "界".repeat(ROOM_APP_MAX_PAYLOAD_BYTES) },
+      })
+    ).toBeNull()
+    expect(decodeRoomAppEnvelope(JSON.stringify({ type: "bogus" }))).toBeNull()
+  })
+
+  it("requires the exact current instance and handshake token", () => {
+    expect(
+      decodeRoomAppClientMessage(
+        {
+          type: "ready",
+          appInstanceId: "shared-canvas:abc123",
+          handshakeToken: "nonce",
+        },
+        "shared-canvas:abc123"
+      )
+    ).toMatchObject({ type: "ready", handshakeToken: "nonce" })
+    expect(
+      decodeRoomAppClientMessage(
+        {
+          type: "sendReliable",
+          appInstanceId: "other:abc123",
+          payload: { type: "stroke" },
+        },
+        "shared-canvas:abc123"
+      )
+    ).toBeNull()
+  })
+
+  it("bounds the participant projection and separates reliable/realtime rate", () => {
+    expect(
+      projectRoomAppParticipants([
+        { participantId: "a", name: "Alice", kind: "human" },
+        { participantId: "b", name: "Pi", kind: "agent" },
+      ])
+    ).toEqual([
+      { participantId: "a", name: "Alice", kind: "human" },
+      { participantId: "b", name: "Pi", kind: "agent" },
+    ])
+    const guard = roomAppRateGuard()
+    for (let index = 0; index < 20; index += 1)
+      expect(guard.allow("reliable", 1, 1000)).toBe(true)
+    expect(guard.allow("reliable", 1, 1000)).toBe(false)
+    expect(guard.allow("realtime", 1, 1000)).toBe(true)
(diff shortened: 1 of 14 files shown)


madawei2699 (commented):
CHANGES REQUESTED on `a0645b68ab937a247c8d2b40115c1e588e92a8b5`.

The overall Phase 0A direction is correct: curated sandboxed iframe, one MessagePort boundary, no per-App PeerConnection, no App payloads in DO/history, client-side guardrails, and a disabled-by-default kill switch all match #354/#344. But I would not merge this head yet.

Blocking findings:

1. **Room App tabs currently self-reset to Room.** The existing stale-Task cleanup effect treats any non-`room` interaction that is not a Task as stale, so `app:<id>` is immediately reset. This is the existing Codex P1 and makes the feature unusable. Exempt valid Room App interactions and add a regression test that an App tab remains mounted.

2. **Optional App channel setup can fail the entire Human Room connection.** `connectMediaSession()` awaits `establishDataChannelTransport()`, and the new App `datachannels/new` branch can throw after the ordinary file channel succeeded. A transient/unsupported App-channel failure must clean up only App channels, disable/fail the App seam locally, and let ordinary chat/voice/files continue. This is also the existing Codex P1 and directly violates #354 failure isolation.

3. **Inbound App messages lose their trusted sender identity.** `subscribeRoomAppChannel()` knows the remote `participant.id`, but `handleRoomAppChannelMessage()` discards it and forwards only `{appInstanceId, lane, payload}`. Both fixtures need a generic way to associate a cursor/unit/op with the participant that actually sent it. Do not make Apps trust a `participantId` supplied inside `payload`; derive a bounded `sourceParticipantId` (name optional/not needed) from the subscribed remote channel and expose that host-generated identity on inbound reliable/realtime messages. Add a test proving payload cannot spoof the sender identity.

4. **The combined byte-rate guard can retain stale bytes from the other lane.** `roomAppRateGuard().allow()` prunes only `events[lane]`, then sums both lane arrays. Old samples in an inactive other lane can therefore keep counting after the 1s window and incorrectly starve the active lane. Prune both lane buckets against the same `windowStart` before computing `totalBytes`, with a cross-lane stale-window regression test.

5. **Unknown App instance IDs are not actually rejected at the transport boundary.** `decodeRoomAppEnvelope()` only regex-validates `appInstanceId`; arbitrary syntactically valid IDs are admitted, rate-counted, and dispatched to listeners, then ignored later by a mounted host. #354 explicitly calls for malformed/unknown App-instance guardrails. Validate inbound IDs against the curated Room App instances for this Room (or an equivalently narrow generic registry) before listener dispatch; do not add a generic extension registry.

6. **The full Cloudflare build is still unverified.** The PR body says local `yarn cf-build` was blocked in the OpenNext/native `sharp` stage, while GitHub `Build & Deploy` is skipped on PRs. Current CI is green for tests/lint/type/docs, but it did not provide the clean full build the PR text expects. Run `yarn cf-build` successfully in a clean environment (or an equivalent existing CI path) before merge.

Please also strengthen the focused transport tests around the code that is currently least exercised: remote Human `appDataChannelReady` -> two remote subscriptions, inbound sender tagging, reliable vs realtime options, session rotation/unmount cleanup, and App-channel failure isolation. The current hook test proves local send only; the remote subscription/retry path is much larger than that coverage.

After these are fixed and the head moves, I can do a narrow re-review. Do not merge yet.

madawei2699 (commented):
CHANGES REQUESTED on `031f68c686a6137a11db11d6f4beb8a0a5476ac6` — one remaining blocker.

The previous six review findings are substantially fixed: App tabs stay mounted; optional local App-channel setup no longer aborts the ordinary Room connection; inbound messages receive host-derived `sourceParticipantId` and spoofed payload fields are rejected; both rate buckets are pruned; unknown App instances are rejected at the transport boundary; and the new remote subscription/retry/rotation/cleanup tests cover the previously missing path. I also resolved the two older Codex inline threads.

Remaining blocker: **a failed local Room App transport can be re-exposed by the subsequent authoritative Room state.** In the App-channel setup catch path, the hook correctly sets `roomAppsEnabledRef.current = false`, `roomAppChannelsReadyRef.current = false`, and `setRoomAppsEnabled(false)`. But after the WebSocket opens, `applyRoomState()` unconditionally does `setRoomAppsEnabled(state.roomAppsEnabled === true)`. The server still advertises the environment-level flag as true, so the UI can show the Room App tabs again even though this browser has no App DataChannels and `sendRoomAppMessage()` will always fail because the local ref remains false. That violates the intended fail-closed/failure-isolated behavior.

Please make the rendered/local capability depend on both the authoritative environment flag and successful local transport readiness (or an equivalent explicit local-disable state), and add a regression test that simulates: session says Room Apps enabled → local App `datachannels/new` fails → WebSocket Room state still says `roomAppsEnabled: true` → `result.current.roomAppsEnabled` remains false and ordinary Room remains healthy.

Build note: I no longer consider the reported local `cf-build` failure a blocker by itself. The failing `src/common/utils.tsx` is byte-for-byte unchanged between the current `cf-sfu` base and this head, and the current base `cf74405f…` completed the repository's actual push `Build` step (`yarn cf-build`) and deploy successfully. PR CI for this head is green for lint/type/docs/tests, with Build & Deploy skipped by the workflow's normal PR condition. Once the one local-capability state bug above is fixed, I expect this to be ready for final approval unless the new head introduces something else.

Housekeeping only: the PR body still says 737 tests and the older `sharp` build explanation; update it to the current 742-test result / current build note when touching the PR, but that is not a merge blocker.

madawei2699 (commented):
APPROVE on `d5e0b75103ac585afcf06e3faa0449231249ef86`.

Narrow re-review of the final capability-state blocker passes.

Verified:
- server-side Room App capability and local App DataChannel readiness are tracked separately;
- exposed `roomAppsEnabled` is true only when both server capability is enabled and local App transport is ready;
- local App-channel setup failure remains fail-closed for the App seam while ordinary Room behavior stays healthy;
- a later authoritative Room state with `roomAppsEnabled=true` no longer resurrects App tabs when local transport setup failed;
- successful setup explicitly enables the local capability;
- regression coverage exercises the full failure -> later Room state path and keeps `roomAppsEnabled=false`;
- prior transport blockers remain fixed: valid App tabs survive stale-task cleanup, optional channel failure isolation, trusted `sourceParticipantId`, cross-lane rate-window pruning, curated instance rejection, and remote subscription/retry/rotation/cleanup coverage;
- exact-head GitHub App Tests and Lint/Type/docs checks pass, and PR Autofix passes.

`Build & Deploy` is expectedly skipped on pull_request by the repository workflow. I do not consider the reported local `utils.tsx` strict-null failure a PR blocker: that file is unchanged by this PR and the current `cf-sfu` baseline successfully ran the same `yarn cf-build` step in its push workflow.

No remaining merge blocker found. Proceed with merge for #354 Phase 0A. After merge, keep #344 open and move to the two Lab fixtures (#12 Shared Canvas and #13 Tiny Arena) against this same seam; do not stabilize a public SDK yet.

Issue #363, opened 2026-09-12
ux(room): make attachments compose-first and align Room actions with Tasks / Room Apps


madawei2699:
## Status

**EVIDENCE-BACKED ROOM UX FOLLOW-UP.**

Recent production dogfood after the Task / Live View and Room App work exposed a small but important interaction mismatch in the current Room UI:

1. selecting a file from the Room composer sends it immediately as a separate timeline item, while the text typed beside it is a separate message;
2. an active Task reuses the text composer but intentionally hides the `+` Room-actions menu, so a Human cannot attach a file from the Task view;
3. the current `+` menu mixes three different concepts — attachment/composer input, an inline Room action (`Poll`), and legacy standalone activities (`Whiteboard`, `Games`) — while Room Apps are becoming a separate top-level interaction surface.

This issue should make the interaction model coherent without turning Free4Chat into a workspace/file product or redesigning Room Apps.

Related completed foundations:

- #164 — Room conversation/composer refresh;
- #234 — ordinary Room message + `@target` + attachments as the primary Human/Agent path;
- #316/#347 — Task + Live View productization/docs;
- #344/#361 — Room App architecture + production dogfood.

Related product experiments:

- #341 — low-cost Room Activities retention hypothesis;
- #346 — measurement guardrails.

Do **not** reopen or broaden those issues for this UX work.

---

### 1. Room attachment selection sends immediately

The Room composer currently exposes:

```text
+
├─ Attach file
├─ Whiteboard
├─ Poll
└─ Games
```

`Attach file` opens the native picker and immediately calls the existing file-send path. The file therefore becomes its own ephemeral DataChannel timeline item; text typed in the composer is sent separately.

That is technically valid but visually/mentally surprising compared with the normal composer model:

```text
select file
→ see pending attachment in composer
→ optionally type caption/instruction
→ press Send once
```

The distinction matters more now that the same composer is the primary Human → Agent interaction surface. A prompt such as:

```text
@Pi please review this screenshot and tell me what is wrong.
[ screenshot.png ]
```

should feel like one Human submission even if the underlying Room still uses separate bounded transport primitives.

### 2. Task view has no Human attachment affordance

`TextChatCard` intentionally hides the `+` menu when `taskRequestId` is present, and `onSendFile` currently accepts only a `File`, not a Task correlation.

This leaves the Task composer text-only from the Human side.

The architecture already supports Task-scoped attachments:

- `/api/room/attachments` accepts `X-Task-Request-Id`;
[...]

PR #365, opened 2026-09-12
ux(room): make attachments compose-first and available in Tasks (#363)

madawei2699:
Implements the **first slice** of #363: **A. compose-first Room attachments** and **B. bounded Human attachments inside an active Task**. `Closes #363`.

**Base:** `93b2f0b91a96fa493ffc55a3f9a09e2b941e0eb2` (`cf-sfu`)
**Head:** `3741dee8603f39a05ebd3df2f5b1dfefe38d40b9` — 14 files, +3512 / −265

Two review rounds are folded in below.

## A1 — compose-first Room attachments

Picking a file creates a **composer draft** (chip with name/size/remove) instead of sending; one Send runs the attachment and the text.

- Send works for text-only, attachment-only, and text + attachment; `@Agent` targeting is resolved from the text exactly as before.
- Enter / Shift+Enter / IME handling is untouched; the Human browser DataChannel path and its **20 MB** limit are unchanged.
- **Readiness edge (review round 1, point 1):** `sendFileMessage(file, { onReadiness })` fires readiness only after the bounded Agent-readable copy has been **published**, when such a copy applies to the file — and it starts that copy **concurrently** with the 20 MB transfer, so the full transfer is never awaited. For a file with no applicable copy, readiness still fires at local initiation. If the copy cannot be published, the composer keeps the draft with a truthful error and does **not** release the text.
- The attachment is initiated first and awaited: a submission whose attachment could not even begin is never completed as an obviously partial text-only send.

## A2 — Human attachments inside an active Task

- The active Task composer exposes a small **Attach** affordance (the Room `+` menu with Whiteboard/Poll/Games stays hidden in Task scope exactly as before).
- Uploads use the existing bounded Room attachment path with `X-Task-Request-Id`; supported image/text types; the existing **768 KB** Agent-readable bound.
- **Derived-copy bound (review round 1, point 2):** for images the 768 KB bound applies to the **derived** `createAgentVisionCopy` result — the same resize/re-encode the ordinary Room path already uses — so a normal multi-MB screenshot now works in a Task too. Text-like sources keep the raw 768 KB bound.
- Renders only in its own Task; Task T never shows Task U's attachment; unknown/expired Task fails closed (409 `unknown_task_request`, 403 when no participating Agent is in the Room, 403 for a non-participating Agent) with no bytes stored and **never** a silent fallback to Room scope.

## Single wake boundary for one Human Task submission (review round 2)

The round-1 fix made every Human Task attachment `addressed: true`, which split `attachment + text` into two Harness turns. The composer now owns the wake intent and the Room **persists** it:

```text
wakeAgent = (text === "")          # chosen per submission by the composer
```

- new `app/src/common/taskAttachmentWake.ts` — a strict `X-Task-Attachment-Wake: "1" | "0"` token; absent/unknown fails closed to *not addressed*;
- `RoomAttachment.taskWake?: boolean` is stored on the existing attachment record (no new schema, store, or transport) and `validAttachment` accepts only `undefined | boolean`, so it survives normalization and a reconnect/replay rebuilds the same `addressed` value;
- `toAttachmentEvent` requires `senderKind === "human" && taskWake === true` **plus** the existing canonical-Task-participation resolution. Agent-authored Task artifacts, Room-scope attachments, and any upload without a validated Task correlation stay unaddressed;
- `room/server.ts` forwards only the two recognized tokens.

Result: **attachment only** → the attachment is the single addressed wake; **attachment + text** → the attachment is persisted first as Task context (unaddressed, lower sequence) and the following Task text is the single addressed wake, so the first Harness turn already sees both.

## Tests

**89 files / 819 tests.** No existing test file was modified or deleted.

- `app/src/room/taskAttachmentWake.test.ts` (new, real Worker route → real `RoomSession`): `attachment + text` yields exactly **one** addressed event **both live and on replay** from retained Room state, with the attachment unaddressed in the same Task scope at a lower sequence and readable by the Task Agent; an idle parked long-poll is **not** resolved by the attachment half; `attachment only` wakes on the attachment (live + replay, asserting the persisted `taskWake === true`); an unresolvable Task still fails closed.
- `common/taskAttachmentWake.test.ts`, `room/server.test.ts`, `do/roomSessionHumanTaskAttachment.test.ts`, `hooks/useSfuChatRoom.taskAttachment.test.tsx`, `components/{TextChatCard,RoomContent}.composeAttachment.test.tsx`: wake-intent encode/parse and per-submission computation, only `"1"`/`"0"` forwarded, `taskRequestId` on the bounded path, derived-copy bound, readiness edge, draft/remove/error semantics, `@Agent` targeting with an attachment present.

## Validation (re-run on this head)

| Command | Result |
|---|---|
| `yarn prettier --check .` | pass |
| `yarn lint --max-warnings 0` | pass, no unintended `--fix` edits |
| `yarn type-check` | pass |
| `yarn docs:check` | pass (docs untouched) |
| `npx vitest run` | **89 files / 819 tests passed** |
| `yarn cf-build` | exit 0, `.open-next/worker.js` produced |
| CI | App Tests / Lint & Type Check / Format & lint autofix all **pass** |

The `Failed to copy <node_modules pkg>` lines during `cf-build` are pre-existing environment noise: the base commit produces the identical set with exit 0.

## Deviations

1. **Server-side changes were unavoidable** — `RoomSession.handleAttachmentUpload` rejected any Human task-correlated upload, and the wake intent has to be persisted server-side to survive replay.
2. `onSendFile` returns `Promise<void> | void` and resolves at genuine local initiation (readiness edge).
3. An **absent** wake token means *not addressed* (fail-closed); the updated composer always sends an explicit token, so this only affects a stale pre-change bundle.
4. `wakeAgent` is a **required** parameter, so every call site must choose.
5. A >20 MB Room pick now errors in the composer chip instead of a transient timeline bubble; the limit itself is unchanged.

## Non-goals honoured

No Room App placement/transport change, no Poll→App conversion, no Whiteboard/Games migration, no new message schema, no permanent storage or public URLs, no threads/replies, no multi-file framework, no analytics expansion, no `agent/**` change, no broad `TextChatCard`/`RoomContent` refactor.

## Remaining concerns

- In a Task composer the `@`-mention picker lists all connected agents; if a Human @-mentions only a **non-participating** agent together with text, no Task-agent wake occurs for that submission. That follows `wakeAgent = text === ""` literally and matches existing Task-text addressing semantics; tightening it (e.g. forcing the canonical Task agent as target) would be a separate change.
- On a readiness-edge copy failure the draft is kept although the Human DataChannel copy may already have transferred, so a retry re-sends the Human file — inherent to the requested fail-closed UX.

Not merged — leaving the merge decision to review.

### DIFF
--- a/app/src/common/taskAttachmentWake.test.ts
+++ b/app/src/common/taskAttachmentWake.test.ts
@@ -1,0 +1,35 @@
+import { describe, expect, it } from "vitest"
+
+import {
+  encodeTaskAttachmentWake,
+  parseTaskAttachmentWake,
+  TASK_ATTACHMENT_WAKE_HEADER,
+} from "./taskAttachmentWake"
+
+/**
+ * #363 second review: the Task attachment wake intent is a bounded boolean
+ * carried on the existing Task-correlated upload and persisted with the
+ * attachment. The transport contract is exactly two one-character tokens so
+ * no client string can smuggle a second wake intent.
+ */
+describe("Task attachment wake intent transport", () => {
+  it("keeps the wire header stable", () => {
+    expect(TASK_ATTACHMENT_WAKE_HEADER).toBe("X-Task-Attachment-Wake")
+  })
+
+  it("round-trips both explicit intents", () => {
+    for (const intent of [true, false]) {
+      expect(parseTaskAttachmentWake(encodeTaskAttachmentWake(intent))).toBe(
+        intent
+      )
+    }
+    expect(encodeTaskAttachmentWake(true)).toBe("1")
+    expect(encodeTaskAttachmentWake(false)).toBe("0")
+  })
+
+  it("never interprets an unknown or absent value as a wake", () => {
+    for (const value of [undefined, null, "", "true", "false", "yes", "2"]) {
+      expect(parseTaskAttachmentWake(value)).toBeUndefined()
+    }
+  })
+})
--- a/app/src/common/taskAttachmentWake.ts
+++ b/app/src/common/taskAttachmentWake.ts
@@ -1,0 +1,31 @@
+// #363 (second review): the ACTIVE-Task composer is compose-first, so one
+// Human Send can carry an attachment and text together. A Task-correlated
+// Human attachment therefore must NOT derive its Agent wake from
+// `senderKind === human && taskRequestId` alone: the composer decides whether
+// this upload is the whole submission and sends that decision as one bounded
+// token on the existing Task-correlated attachment upload. The Room persists
+// it with the attachment record, so a reconnect/replay rebuilds the same
+// `addressed` value — an immediate-only broadcast would not survive replay.
+//
+//   "1" => attachment-only submission: the persisted attachment addresses the
+//          canonical participating Task Agent(s) and wakes an idle Harness.
+//   "0" => attachment + text submission: the attachment is Task context only,
+//          and the following addressed Task text is the single wake boundary.
+//
+// Anything else — including an absent header — is "no explicit wake intent"
+// and never wakes an Agent on its own.
+export const TASK_ATTACHMENT_WAKE_HEADER = "X-Task-Attachment-Wake"
+
+export function encodeTaskAttachmentWake(wakeAgent: boolean): "1" | "0" {
+  return wakeAgent ? "1" : "0"
+}
+
+/** Strict, bounded parse: only the two exact one-character tokens are
+ * accepted, so no client string can smuggle a second wake intent. */
+export function parseTaskAttachmentWake(
+  value: string | null | undefined
+): boolean | undefined {
+  if (value === "1") return true
+  if (value === "0") return false
+  return undefined
+}
--- a/app/src/components/RoomContent.composeAttachment.test.tsx
+++ b/app/src/components/RoomContent.composeAttachment.test.tsx
@@ -349,12 +349,37 @@ describe("Active Task attachment path (#363 A2)", () => {
     fireEvent.click(screen.getByLabelText("Send message"))
 
     await waitFor(() =>
-      expect(sendTaskAttachment).toHaveBeenCalledWith(file, "T")
+      expect(sendTaskAttachment).toHaveBeenCalledWith(file, "T", false)
     )
     expect(sendFileMessage).not.toHaveBeenCalled()
     await waitFor(() =>
       expect(sendTextMessage).toHaveBeenCalledWith("use this", [], "T")
     )
+  })
+
+  it("asks the Room to wake the Task Agent for an attachment-only submission", async () => {
+    const sendFileMessage = vi.fn()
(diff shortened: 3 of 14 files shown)


madawei2699 (commented):
Second-pass review on `836ff5d886c973712d3ff401b75fa1c3a10f24d2`: the three previous blockers are substantially addressed, but there is one new correctness blocker in the combined Task attachment + text path.

Current order in `TextChatCard.sendCurrentMessage()` is:

1. `await onSendTaskFile(attachment, taskRequestId)`
2. then `onSendText(text, targets, taskRequestId)`

The new server behavior makes every Human Task attachment itself `addressed: true` for participating Task Agents. `handleAttachmentUpload()` broadcasts/resolves Agent waiters before the POST returns. Therefore a Human submission like:

```
[ screenshot.png ]
please inspect this
```

can wake the Harness on the attachment event before the accompanying text has even been persisted. The text then becomes a second addressed Task turn. That recreates the same split-submission race this PR is trying to eliminate, just inside Task scope.

Please preserve both desired behaviors:

- attachment-only Task submit => attachment is addressed and wakes the Task Agent;
- attachment + text Task submit => attachment is persisted first as Task context but does NOT independently wake; the following Task text is the single addressed wake boundary, so the first Harness turn sees both attachment + instruction.

The narrowest shape is likely to make the Task upload carry/persist an explicit bounded wake/addressed intent chosen by the composer (`text === ""`), rather than deriving `addressed` unconditionally from `senderKind === human && taskRequestId`. The replayed Agent event must preserve that choice, so avoid an immediate-only wake that cannot be reconstructed after reconnect.

Add an end-to-end-ish regression that proves a Task `attachment + text` composer submission yields one wake/turn boundary with the attachment visible in that turn, while attachment-only still wakes. Everything else in this revision looks consistent with the prior review.

madawei2699 (commented):
Final review on `3741dee8603f39a05ebd3df2f5b1dfefe38d40b9`: **APPROVE / safe to merge**.

The second-review blocker is closed. The Task composer now persists an explicit bounded wake intent: attachment-only submissions wake via the attachment, while attachment+text persists the attachment first as unaddressed Task context and uses the following Task text as the single addressed wake boundary. The real Worker-route → RoomSession regression also verifies replay reconstructs the same boundary and that the attachment is readable inside that first turn. The prior readiness/image-bound/fail-closed fixes remain intact.

I do not see a code blocker on this head. Current `cf-sfu` advanced by one unrelated `app/wrangler.jsonc` Room Apps enablement commit; because this PR does not touch that file, GitHub reports it mergeable, and the head CI is green, I would **not rebase solely for that commit**.

One non-blocking PR-description correction: the body says “No existing test file was modified or deleted,” but `app/src/room/server.test.ts` is an existing test file that was extended. No existing tests were deleted; the wording should just be corrected for the historical record.

Issue #364, opened 2026-09-12
runtime(task): reduce turn stalls and Live View authoring friction after production dogfood


madawei2699:
## Status

**EVIDENCE-BACKED TASK / RESIDENT-RUNTIME FOLLOW-UP. Priority: high.**

A real production Task + Pi dogfood run exposed a compound failure mode where a simple Live View request took ~22 minutes wall-clock even though the eventual successful authoring/publish path required only about a minute once the contract was understood.

This is not a request to expand the Live View component model. The visual limits of the current DSL are separate.

The problems here are:

1. the exact current Live View authoring contract is not cheaply discoverable from the installed Runtime binary, so the Harness spent multiple turns reverse-engineering it;
2. the default ACP turn deadline killed two attempts at about two minutes each;
3. failed addressed work remained eligible for at-least-once retry, but the observed retry did not happen promptly and produced a long silent gap;
4. user-facing assistant-message chunks included work narration/tool-discovery preambles that were later published into the Room reply;
5. strict FIFO made follow-up messages look stale/repetitive after the long stall.

Do not solve this with a scheduler, semantic message reordering, a bigger Live View framework, or a new task system.

---

## Production evidence

The observed run, reduced to non-sensitive timing/behavior:

```text
Task request arrives
→ Agent accepts

attempt 1
~11 tool steps
→ searches local config / binary strings / source / tests to infer Live View schema
→ ACP turn aborts at ~2 minutes
→ turn_failed

~16 minute idle gap
(no useful tool work)

same pending Task work is delivered again

attempt 2
~14 tool steps
→ repeats schema investigation
→ eventually finds the existing Tasks + Live Views docs
→ ACP turn aborts again at ~2 minutes
→ turn_failed

attempt 3
schema now understood
→ write JSON
→ local validation
→ publish Live View
→ attach artifact
→ collab completed
~1 minute
```

So the ~22 minute user-visible delay was mostly:

```text
~4 minutes failed contract discovery
+ ~16 minutes retry gap
+ ~1 minute successful work
```

Two `turn_failed` events were observed. The third attempt succeeded. Do not describe this as a single 20-minute turn timeout.

The same dogfood also showed:

- short English work preambles such as schema/tool-search narration appearing in the eventual Room reply;
- Human follow-ups sent while the first turn was busy being delivered later in strict order;
- repeated Human questions being answered separately because they were distinct canonical Room messages.

The last two points are important context, but strict FIFO itself is not proven wrong.

---

[...]

PR #366, opened 2026-09-12
runtime(task): make Live View authoring discoverable and recover failed turns (#364)

madawei2699:
Implements #364: **runtime turn recovery** — bounded retry for an ambiguous turn, plus honest reporting instead of silent loss. `Refs #364`.

**Base:** `93b2f0b91a96fa493ffc55a3f9a09e2b941e0eb2` (`cf-sfu`)
**Head:** `661df9eec0c16c718ddd1c20fb0c82752d983305` — 12 files, +2327 / −113

All changes are inside `agent/**`. Two review rounds are folded in below.

## Defect

`RunTurn` success is the **only** acknowledgement boundary. A turn that fails or is ambiguous after the prompt was already delivered to the Harness stayed unacknowledged (`pendingAddressed`, `deliveredThrough`) and the next `drainTurns` simply re-read the same Room state — so an ambiguous turn was either lost forever or re-delivered unpredictably, with the Human given no truthful signal.

## What changed

**A. Bounded retry.** `maxTurnRetryAttempts = 2` (**3 attempts total**, never more), `RetryDelay` 1s → 2s. Only *transient* failures retry: transport/send, session/Harness-process replacement, and timeout. Permanent failures (validation, `agent_error`, cancelled, conflict) do not. The timeout stays at the existing **120 s** `defaultTurnTimeoutMs`.

**B. Truthful terminal reporting.** A turn that exhausts its attempts becomes a **parked** scope, reported through the existing bounded diagnostics vocabulary (`scopeKind`, `failureClass` ∈ `timeout|process|session|send|other`, `elapsedMs`, `retryAttempt`, `retryDelayMs`). The Room scope never falls back to Task scope or vice versa, and the diagnostics carry no Human content.

**C. Parked scope is inert until re-addressed.** `closedTurnRecovery map[canonicalTurnKey]struct{}`; `nextRunnableTurn()` skips parked scopes so unrelated Room traffic can **never** re-execute them, parked scopes do not block other scopes, and re-arming happens only through a newly accepted addressed trigger for that same scope (`reopenTurnRecoveryLocked`).

**D. State.** An exhausted turn no longer leaves the Agent in `StateReconnecting`; a parked scope reports the honest waiting state.

### Review round 2 — `adoptJoin` falsy-vs-nil (this is the narrow blocker that was outstanding)

`closedTurnRecovery` was tested for **non-nilness**, not for **content**:

```go
// before
if r.closedTurnRecovery != nil { ... }   // any allocated-but-empty map ⇒ false "reconnecting"
// after
if r.unresolvedTurnFailureLocked() { ... }
```

An empty-but-non-nil `closedTurnRecovery` (trivially reachable via `reopenTurnRecoveryLocked` deleting the last entry, or a map allocated for a scope that then succeeded) made a **rejoin** report `connecting`/`reconnecting` when nothing was actually unresolved. `unresolvedTurnFailureLocked()` is now the single truthful-state predicate, used by all three consumers — the `drainTurns` deferred restore, `restoreStateAfterRetry`, and `adoptJoin` — so state derivation can no longer disagree with itself. Regression test `TestExplicitReAddressThenSuccessLeavesNoStaleClosedRecovery`, confirmed to fail against the old predicate (`turn_retry_test.go:534: transport rejoin after recovery reported State:reconnecting, want waiting`).

### Adjacent fixes in the same invariant

- **`live-view describe` shares the validator's vocabulary.** `liveViewComponentFields`, `liveViewComponentRequiredFields`, `liveViewActionTypes`, `liveViewActionFields`, `minLiveViewRevision`, and `liveViewIncrementBound` are now used by both `validateTaskLiveViewJSON` and the CLI, so `describe` can no longer advertise a field the validator rejects. Limits unchanged: 32 KB JSON, 64 components, depth 8, 32 data keys, 400-char text, 80-char labels, increment ±1000.
- **`voiceOutput()` `mediaMu` race** kept as previously fixed.
- **ACP/prompt handling** in `harness/` aligned with the same session-replacement semantics.

## Tests

`turn_retry_test.go` plus `runtime`, `cli`, `harness` coverage: bounded attempt count is asserted exactly (never a 4th attempt), permanent failures do not retry, timeout classification, exhausted turn parks and is **not** re-executed by unrelated Room events, parked scope does not block other scopes, only an accepted addressed trigger re-arms, success-only acknowledgement, and the round-2 regression above. `-race` clean on `internal/runtime` and `internal/voice`.

## Validation (re-run on this head)

| Command | Result |
|---|---|
| `gofmt -l .` | clean |
| `go vet ./...` | clean |
| `go test ./... -count=1` | **12 packages pass** |
| `go test -race ./internal/voice ./internal/runtime` | pass |
| `go build ./cmd/free4chat-agent` | pass |
| `agent/scripts/check-cleanup.sh` | pass |
| CI | gofmt/vet/test/build, Distribution/cleanup, 4 platform builds all **pass** |

## Non-goals honoured

No scheduler, background goroutine, or busy loop; no unbounded retry; no change to the 120 s timeout; no Human content in diagnostics; no cross-scope fallback; no production version pin or release change; no `app/**` change.

## Remaining concerns

- `lastError` is a single global slot: a later transport error can replace a parked turn's error text while the state correctly remains a waiting state. The **state** is already truthful; making the text per-scope is a follow-up.
- A parked scope waits for an addressed trigger for that scope — deliberate FIFO, but it means a Human must re-address rather than the Runtime self-healing.
- Harness session/process replacement does not by itself re-arm a parked turn; only a new addressed trigger does.
- `harnessFailed` is never set to `true` — dead field, cosmetic.

## Dogfood

Production Pi fresh-resident dogfood is required after merge/release, and #364 should stay open until it passes; that is a post-merge step, not a code blocker.

Not merged — leaving the merge decision to review.

### DIFF
--- a/agent/internal/cli/cli.go
+++ b/agent/internal/cli/cli.go
@@ -56,6 +56,7 @@ func usageText() string {
   free4chat-agent surface publish --file <snapshot.jpeg|png|webp> [--instance <id>]
   free4chat-agent surface clear [--instance <id>]
   free4chat-agent surface read --participant <participant-id> [--instance <id>]
+  free4chat-agent live-view describe [--json]
   free4chat-agent live-view publish --task-request-id <id> --file <surface.json> [--instance <id>]
   free4chat-agent context read [--before-sequence <n>] [--after-sequence <n>] [--limit <1-50>] [--before-transcript-sequence <n>] [--after-transcript-sequence <n>] [--transcript-limit <1-50>] [--instance <id>]
   free4chat-agent version [--json]
@@ -473,31 +474,40 @@ func run(args []string) error {
 		}
 
 	case "live-view":
-		if len(rest) == 0 || rest[0] != "publish" {
+		if len(rest) == 0 {
 			return errUsage()
 		}
-		filePath := option(rest[1:], "--file")
-		taskRequestID := option(rest[1:], "--task-request-id")
-		if filePath == "" || taskRequestID == "" {
+		switch rest[0] {
+		case "describe":
+			return runLiveViewDescribe(rest[1:])
+		case "publish":
+			// Behaviour is unchanged: describe only adds a pure local,
+			// machine-readable authoring contract (issue #364 A).
+			filePath := option(rest[1:], "--file")
+			taskRequestID := option(rest[1:], "--task-request-id")
+			if filePath == "" || taskRequestID == "" {
+				return errUsage()
+			}
+			data, err := attachments.ReadBounded(filePath, maxTaskLiveViewBytes)
+			if err != nil {
+				return fmt.Errorf("Live View JSON must be a non-empty file up to %d bytes", maxTaskLiveViewBytes)
+			}
+			var surface map[string]any
+			if err := json.Unmarshal(data, &surface); err != nil || len(surface) == 0 {
+				return errors.New("Live View JSON must be an object")
+			}
+			if err := validateTaskLiveViewJSON(data); err != nil {
+				return err
+			}
+			return runViaDaemon(&daemon.IpcRequest{
+				Op:            "live-view-publish",
+				InstanceID:    option(rest[1:], "--instance"),
+				TaskRequestID: taskRequestID,
+				Surface:       surface,
+			})
+		default:
 			return errUsage()
 		}
-		data, err := attachments.ReadBounded(filePath, maxTaskLiveViewBytes)
-		if err != nil {
-			return fmt.Errorf("Live View JSON must be a non-empty file up to %d bytes", maxTaskLiveViewBytes)
-		}
-		var surface map[string]any
-		if err := json.Unmarshal(data, &surface); err != nil || len(surface) == 0 {
-			return errors.New("Live View JSON must be an object")
-		}
-		if err := validateTaskLiveViewJSON(data); err != nil {
-			return err
-		}
-		return runViaDaemon(&daemon.IpcRequest{
-			Op:            "live-view-publish",
-			InstanceID:    option(rest[1:], "--instance"),
-			TaskRequestID: taskRequestID,
-			Surface:       surface,
-		})
 
 	case "readiness":
 		return runReadiness(rest)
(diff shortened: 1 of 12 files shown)


madawei2699 (commented):
Second-pass review on `94ec3d8efa3674378d7031d39d9f31a3b09f8573`: the previous bounded-retry and permanent-failure blockers are fixed correctly, but I found one small state bug in the new closed-recovery lifecycle.

`adoptJoin()` currently tests:

```go
if r.closedTurnRecovery != nil {
    r.state = StateReconnecting
} else {
    r.state = StateWaiting
    r.lastError = ""
    r.lastErrorSource = ""
}
```

But `reopenTurnRecoveryLocked()` and `forgetTurnRecoveryLocked()` delete entries without setting the map back to nil. So after:

1. a turn exhausts and creates `closedTurnRecovery`;
2. a new addressed trigger explicitly reopens it;
3. the old turn succeeds and the last closed marker is deleted;
4. a later transport rejoin occurs;

the map is empty but still non-nil, and `adoptJoin()` incorrectly reports `StateReconnecting` instead of a healthy waiting state. This is exactly the kind of stale truthful-state bug the new predicate is trying to prevent.

Please use the actual unresolved condition there (at minimum `len(r.closedTurnRecovery) > 0`; ideally the same locked predicate if that matches the intended join semantics) and add a regression covering: exhaust/park -> explicit same-scope re-address -> successful recovery -> marker count zero -> transport rejoin -> `StateWaiting` with no stale Harness error.

The rest of the new park/reopen logic now looks coherent: unrelated Room events no longer execute an exhausted turn, another scope can progress, retryAttempt stays bounded, and permanent failures remain pinned/reconnecting. Keep the `voiceOutput()` race fix.

madawei2699 (commented):
Final review on `661df9eec0c16c718ddd1c20fb0c82752d983305`: **APPROVE / safe to merge** for the code.

The remaining state-machine blocker is closed: `adoptJoin`, drain restoration, and transport retry now share `unresolvedTurnFailureLocked()`, so an empty-but-non-nil `closedTurnRecovery` cannot resurrect a stale reconnecting state. The new exhaust → re-address → successful recovery → transport rejoin regression covers the exact failure mode. The bounded parked-turn behavior, cross-scope progress, retry budget, acknowledgement boundary, Live View descriptor, reply-hygiene contract, and the `voiceOutput()` race fix all still look correct.

I would **not rebase solely for current `cf-sfu` `9598227c8`**: that base advance is one unrelated `app/wrangler.jsonc` Room Apps enablement commit, while this PR is entirely `agent/**`; GitHub reports mergeable and the current head CI is green.

Before merge, please correct the PR body only (no code/CI rerun needed):
1. “parked scope reports the honest waiting state” is stale — the implementation/tests intentionally keep genuinely parked unresolved work in `StateReconnecting` with `LastError`; only after successful recovery does it return to `StateWaiting`.
2. Post-cognition Room send failures do **not** retry Harness cognition by design, so the sentence saying “transport/send” failures retry is too broad.
3. `permanentTurnFailure` currently special-cases the deterministic scoped-Harness-unsupported case; avoid claiming that validation/`agent_error`/cancelled/conflict are all explicitly classified as permanent unless the code actually does so.

These are PR-record accuracy fixes, not code blockers. Keep `Refs #364`; production fresh-Pi dogfood after release remains the close gate for #364.

PR #369, opened 2026-09-12
ux(room): put active Room Apps on the visual Stage (#367)

madawei2699:
Closes #367
Refs #361
Refs #344

Production dogfood showed that selecting Shared Canvas / Tiny Arena replaced the right-hand Room/Task conversation pane: the large left Stage stayed idle, the App was cramped into the conversation column, and Room chat disappeared. This is the narrow placement correction — the #356 bridge, transport, sandbox, catalog and Lab fixtures are all untouched.

## State model

One `app:<id>` interaction value used to imply both "which App is open" and "which conversation is selected". They are now two independent selections:

```text
conversation scope (existing interaction tablist, unchanged)
  room | taskRequestId

visual Stage (new, narrow)
  participants | screen | live view | active Room App
```

- `activeInteraction` only ever holds `room` or a Task request id; the `app:*` values and the App tabs in the right pane are gone.
- `activeRoomAppId` (existing state) now independently owns "an App is open on the Stage", resolved against the curated catalog only.
- An open App survives conversation switching: Room ↔ Task no longer unmounts the iframe or closes its MessagePort. Only `Close`, or the catalog losing the App (e.g. `ROOM_APPS_ENABLED` off), tears it down.
- Screen / Live View remain the same `stageView` preference they were; selecting either also dismisses an open App, so the Stage always shows one thing.

## Behavior

Desktop:

```text
┌───────────────────────────────┬──────────────────────────┐
│ Shared Canvas / Arena /       │ Room | Task A | Task B   │
│ screen share / Live View /    │                          │
│ participants (Stage, 75%)     │ conversation + composer  │
└───────────────────────────────┴──────────────────────────┘
```

- The App mounts in the existing left Stage seam through the unchanged `RoomAppHost`; the right pane always renders `TextChatCard` for the selected Room/Task conversation.
- An active App gets the wide Stage (75%), the same treatment screen share already had, instead of a conversation-sized pane.
- Entry: a small `Stage` switcher row with the two curated catalog Apps plus the existing Screen / Live View buttons when both exist. No registry, marketplace, manifest, SDK, or generic navigation — it is the same curated catalog, rendered as buttons.
- Close returns the Stage to the participant grid, with the conversation scope untouched.

Mobile: no new navigation framework. The App is Stage content rather than conversation content, so in the existing stacked responsive layout it owns the full-width Stage pane instead of the conversation pane; Room/Task conversation stays reachable below it exactly as before. Real iOS/touch smoke remains under #361.

## Changed files

[...]

Issue #361, opened 2026-09-11
room-app phase 0b: deploy merged Lab fixtures and run production dogfood


madawei2699:
## Status

**READY — this is the only active Room App execution issue after Phase 0A/static fixture proof.**

Parent: #344  
Lab fixtures: `i365dev/free4chat-extension-lab#12`, `#13`

Phase 0A is complete:

- #356 merged the generic sandbox host + bounded reliable/realtime bridge;
- #354 is closed;
- Lab Shared Canvas PR #14 is merged;
- Lab Tiny Arena PR #15 is merged;
- both materially different fixtures consumed the same #356 seam without adding Canvas/Arena domain logic to Free4Chat core.

The remaining Phase 0 question is runtime reality, not more abstraction:

> **Can the merged fixtures be deployed and used inside real Free4Chat Rooms across desktop/mobile participants while ordinary Room collaboration stays healthy and realtime cost remains bounded?**

## Current production gap

Free4Chat's curated production catalog already points to:

```text
https://room-apps.free4.chat/shared-canvas
https://room-apps.free4.chat/tiny-arena
```

but production currently keeps:

```text
ROOM_APPS_ENABLED=false
```

The Lab repository verifies both fixtures but does not yet deploy them together under the curated production origin/path contract.

### A. Lab deployment

In `i365dev/free4chat-extension-lab`, implement the smallest deployment shape that serves both merged fixtures at the existing curated paths:

```text
room-apps.free4.chat/shared-canvas
room-apps.free4.chat/tiny-arena
```

Requirements:

- use the already merged fixture code;
- keep Canvas/Arena domain semantics inside the Lab;
- do not create a public SDK/manifest/registry;
- do not add a new realtime backend;
- do not expose arbitrary user-provided iframe URLs;
- preserve the exact #356 MessagePort/bridge contract;
- prefer one thin deployment surface for the two paths if that is simpler than independent workers;
- keep deployment reproducible through repo CI/config rather than manual undocumented state.

### B. Free4Chat activation

After both HTTPS paths are live and independently smoke-tested, make the smallest Free4Chat change/configuration required to enable the existing Room App experiment.

Do not redesign the host. The existing kill switch must remain easy to disable.

If activation can be done safely without a code change, document exactly how. If a repo change is required, use one narrow PR.

### C. Production dogfood

Run real Room tests through the production SFU/DataChannel path.

Shared cases:

```text
2 Humans
→ open App
→ ordinary chat + voice remain healthy
[...]

PR #372, opened 2026-09-12
ux(room-app): keep launched App sessions resident across Stage switches (#371)

### LINKED ISSUE #371: ux(room-app): keep launched App sessions resident across Stage switches and separate hide from reset
## Status

**PRODUCTION-DOGFOOD FOLLOW-UP / ROOM APP LIFECYCLE REVISION.**

Production dogfood under #361 after #369 validated the Stage placement, but exposed that the current host lifecycle still treats visual navigation as App-session destruction.

Related:
- #344 — Room App experiment umbrella
- #361 — production dogfood
- #367 / PR #369 — completed Stage placement correction
- `i365dev/free4chat-extension-lab#12` — Shared Canvas
- `i365dev/free4chat-extension-lab#13` — Tiny Arena

## Production evidence

The new Stage UX is working well:

- Shared Canvas renders on the large Stage while Room/Task conversation remains visible;
- Tiny Arena renders on the large Stage while conversation remains visible;
- Screen Share and Live View remain reachable as sibling Stage surfaces;
- ordinary Room/Task/attachment/screen-share interaction remained healthy in the same dogfood session.

However, switching visual surfaces currently unmounts the active `RoomAppHost`.

Observed consequences:

```text
Canvas open
→ draw state exists locally
→ switch to Arena / Screen / Live View
→ Canvas iframe + MessagePort are destroyed
→ switch back
→ new iframe starts from empty local state
```

Shared Canvas can recover if at least one other participant still has an incumbent Canvas session mounted, because its App-level reliable bootstrap asks an incumbent for the bounded canonical state. If **all** participants have switched away / closed the App, every in-browser replica is destroyed and reopening starts empty.

Tiny Arena similarly loses its local session on Stage switching. Reopening creates a new epoch and can temporarily show stale/spawn state until current transforms arrive; this fixture also has a separate App-local first-open/reopen convergence follow-up under Lab #13.

This is not a request for server persistence. It is a mismatch between **visual Stage visibility** and **Room App session lifetime**.

## Product invariant

A Room App should behave like a Room-scoped collaborative capability, not like a disposable tab render.

Separate three concepts:

```text
App instance lifetime
→ scoped to the current browser's Room session after first launch

Stage visibility
→ which surface is currently visible

App content reset/new round
→ owned explicitly by the App's domain UI
```

Therefore:

> Switching Stage surfaces, switching between Apps, or visually closing/hiding an App must not implicitly reset its collaborative state.

For Canvas, reset is already explicit (`Clear canvas`).
For a future game, `Ready`, `Restart`, `New round`, etc. should be App-owned semantics, not inferred from host navigation.

## Desired narrow host behavior

Keep the current small curated catalog and #356 bridge.

When a Human launches an App for the first time:

```text
launch App
→ mount RoomAppHost / iframe / MessagePort
→ add it to the browser-local set of launched Room Apps
→ show it on Stage
```

After that, navigation should be visibility-only:

```text
Canvas → Arena
→ Canvas host stays mounted but hidden/non-interactive
→ Arena host becomes visible

Arena → Screen Share / Live View / participants
→ Arena host stays mounted but hidden/non-interactive

Close/Hide
→ remove App from Stage
→ keep its launched session resident
```

Reopening the same App during the same browser Room session should reuse the **same iframe / MessagePort / App-local state**, not create a new session epoch.

The host may continue to cap resident Apps to the existing small curated/max-instance bound. Do not generalize into a workspace/window manager.

## Cleanup boundary

A resident App host should actually be destroyed when its Room/browser lifetime ends, for example:

- leaving the Room;
- browser/page unload;
- `ROOM_APPS_ENABLED` becomes false / catalog removes the App;
- an explicit future destructive App-session action if one is ever introduced.

Do **not** add Durable Object App state, Room history state, database persistence, or external App persistence to solve this issue.

## Resource / cost discipline

Hidden Apps remain client-side and continue to receive the existing bounded logical App messages so they can preserve/reconcile state.

Requirements:

- hidden iframe must not capture keyboard/pointer/focus;
- do not add heartbeat traffic merely to keep it alive;
- preserve existing payload/rate bounds;
- no new PeerConnection;
- no high-frequency DO/history/storage path;
- if a hidden fixture has an animation loop, verify Stage hiding does not cause unexpected network traffic; a domain App may later add its own pause/resume optimization if evidence requires it.

Do not add a new host protocol (`suspend`/`resume`) unless actual production evidence proves simple resident-hidden mounting is insufficient.

## State model

Conceptually:

```text
activeConversation
= room | taskRequestId

activeStage
= participants | screen | liveView | roomApp:<id>

launchedRoomApps
= browser-local bounded set of App ids whose hosts remain mounted
```

Exact variable names are implementation details.

Important:

- `activeRoomAppId` alone must no longer determine whether an App host exists;
- switching Canvas ↔ Arena must not close the previous MessagePort;
- selecting Screen / Live View must not close launched App MessagePorts;
- Room ↔ Task conversation switching remains independent as established by #369.

## Acceptance

- [ ] first Canvas launch creates one iframe/MessagePort;
- [ ] Canvas → Arena hides Canvas but does not unmount it or close its MessagePort;
- [ ] Arena → Canvas restores the exact same Canvas iframe instance and local state;
- [ ] Stage switch to Screen or Live View preserves launched App hosts;
- [ ] visual Close/Hide removes the App from Stage but reopening restores the same host/session;
- [ ] if all Humans hide Canvas and later reopen it, previously retained Canvas state still exists while their Room browser sessions remained connected;
- [ ] switching Room ↔ Task conversation does not affect resident App sessions;
- [ ] disabling Room Apps/catalog removal tears resident hosts down cleanly;
- [ ] leaving/unmounting RoomContent closes each resident MessagePort exactly once;
- [ ] no hidden iframe receives keyboard/pointer interaction;
- [ ] ordinary Room text/voice, Task/Live View and Screen Share remain green;
- [ ] existing #356 protocol and transport tests remain green;
- [ ] no server-side App persistence or new realtime backend is added.

## Non-goals

- persistence after every participant leaves/reloads the Room;
- Durable Object / database App snapshots;
- public Room App SDK;
- manifest/registry/marketplace;
- arbitrary URLs;
- generic tabs/windows/workspace framework;
- gameplay/reset semantics in Free4Chat core;
- Agent semantic App API;
- fixing Tiny Arena's App-local first-open current-transform synchronization (keep that fixture-local under Lab #13).

## Exit

Ship the narrow resident-session lifecycle correction directly into production dogfood, then rerun the #361 Canvas/Arena close/switch/reopen cases.

If this works, Stage navigation becomes presentation-only while App state semantics remain owned by the App — preserving the thin-core boundary.

madawei2699:
Closes #371
Refs #361
Refs #344

Production dogfood after #369 confirmed the Stage placement, then exposed the remaining host-lifecycle bug: Stage navigation unmounted `RoomAppHost`, so hiding Canvas for Arena / Screen / Live View destroyed its iframe, MessagePort and App-local state. Reopening started empty, and once every Human had switched away, no in-browser replica was left to serve Canvas's canonical bootstrap.

This makes App session lifetime a browser-Room-scoped concern and Stage navigation presentation-only. No protocol, transport, catalog, Lab or persistence change.

## State model

```text
activeConversation   room | taskRequestId                    (unchanged, #369)
activeStage          participants | screen | liveView | roomApp
                     (which App is *visible*)
launchedRoomApps     bounded browser-local set of App ids whose hosts stay mounted
```

- `activeRoomAppId` no longer decides whether a host exists. The first Stage entry click launches: mount `RoomAppHost`, add the App to `launchedRoomApps`, show it. `launchedRoomApps` is bounded by the existing `ROOM_APP_MAX_INSTANCES` and can never exceed the curated catalog the panel already slices.
- Render: one visibility slot per resident App (`room-app-slot-<id>`), each holding an unkeyed-by-visibility `RoomAppHost`. The active one gets `flex min-h-0 flex-1 flex-col`; the rest get `hidden` plus `inert` and `aria-hidden="true"`.
- Hide paths (Stage entry toggled off, the host's own Close button, switching to Screen / Live View) only clear `activeRoomAppId`. No unmount, no port close, no re-bootstrap, no new epoch.
- Re-showing a resident App renders the same `RoomAppHost` at the same position in the list, so React reuses the exact iframe element and MessagePort.
- Teardown boundary: the catalog losing an entry — which covers `ROOM_APPS_ENABLED` turning off — prunes `launchedRoomApps` and unmounts the host; unmounting `RoomContent` (leaving the Room) unmounts them all. Each `RoomAppHost` closes its own port exactly once in its existing cleanup.
- No generic reset was added. Per-App reset stays App-owned: Canvas already has *Clear canvas*; a future game owns Ready/Restart.

## Hidden Apps: alive, but not interactive

Hidden slots are `display: none` (out of hit-testing and tab order) **and** `inert` + `aria-hidden="true"` (explicitly non-focusable and out of the accessibility tree). For browsers older than `inert` support, `display: none` alone already provides the non-interactivity guarantee.

Hidden hosts stay mounted and keep receiving the bounded logical App messages they already received, so a hidden Canvas remains a valid canonical-bootstrap incumbent and can reconcile. Nothing new is sent to keep them alive: visibility changes produce no host traffic, no heartbeat, no extra PeerConnection, and no DO/history/storage write. A hidden fixture with an animation loop simply keeps running client-side; any pause/resume optimisation stays a domain decision under Lab #13.

## Tests

`RoomContent.test.tsx` — the Stage describe block is now residency-focused (13 cases). The five #369 cases that asserted destroy-on-switch were replaced by their residency equivalents, the two Stage-entry availability cases from the #369 review are preserved, and the following were added:

1. first Canvas launch creates exactly one iframe and one MessagePort, and the Stage takes 75% while Room chat stays selected and rendered;
2. launch once → hide → show reuses the same iframe and the same port (no new channel, no outbound traffic);
3. visual Close is Hide: the Stage returns to participants, the host stays mounted, reopening restores the identical iframe/port;
4. Canvas → Arena leaves the Canvas iframe and port alive behind the visible Arena, and returning to Canvas restores the exact same Canvas iframe;
5. two RoomContent instances (Alice + Bob) both launch Canvas, both hide it, both reopen: neither resident session was destroyed and each gets its own iframe back — the production "both Humans hid Canvas" case;
6. a hidden App still receives a remote reliable payload through the resident port, and hiding produces no `send` traffic;
7. App → Screen / Live View hides the App, keeps the host and port, and leaves the Task conversation scope selected;
8. Screen-only and LiveView-only Stage entries stay reachable while an App is open (the #369 review cases, updated to residency assertions);
9. hidden slots are `hidden` + `inert` + `aria-hidden="true"`, the visible slot is not;
10. catalog removal (`ROOM_APPS_ENABLED` off) unmounts both resident hosts and closes both ports exactly once;
11. unmounting `RoomContent` closes every resident MessagePort exactly once;
12. Room ↔ Task conversation switching keeps the same resident host (from #369);
13. first-launch Stage placement (from #369).

Mutation-checked: rendering only the visible App (the old destroy-on-switch shape) fails 8 residency cases; dropping `inert`/`aria-hidden` fails exactly the non-interactive case.

`git diff -w` on `RoomContent.tsx` is 94 lines of real change; the raw diff is larger only because the existing Screen / Live View / participants branches moved one JSX level deeper.

## Verification

```
prettier --check .                       pass
yarn lint --max-warnings 0               pass
yarn type-check                          pass
yarn docs:check                          pass
yarn test                                831 passed / 89 files
yarn cf-build                            pass
git diff --check                         clean
```

## Non-goals / not changed

No Durable Object, Room history, database or external App persistence; no new host protocol (`suspend`/`resume`); no heartbeat; no new PeerConnection or realtime backend; no arbitrary URLs, SDK, registry, manifest, workspace/window manager or generic reset; #356 protocol, MessagePort handshake, iframe sandbox, catalog URLs, reliable/realtime lanes, payload/rate guards, SFU, Task Live View, Screen Share, #363 composer/attachments and the Lab fixtures are untouched. Tiny Arena's App-local first-open/reopen current-transform convergence is **not** touched here and remains Lab #13.

### DIFF
--- a/app/src/components/RoomContent.test.tsx
+++ b/app/src/components/RoomContent.test.tsx
@@ -29,6 +29,7 @@ import type { Message } from "@common/types"
 import { trackAnalyticsEvent } from "@common/utils"
 
 import RoomContent from "./RoomContent"
+import { roomAppInstanceId } from "../common/roomApp"
 import { RoomSession } from "../do/RoomSession"
 import type { RoomRecord, RoomState } from "../room/types"
 
@@ -1126,12 +1127,16 @@ describe("RoomContent — Turnstile widget lifecycle", () => {
     expect(screen.queryByText("Connect local Runtime")).not.toBeInTheDocument()
   })
 
-  describe("Room App Stage placement", () => {
+  describe("Room App Stage placement and residency", () => {
     class TestPort {
       onmessage: ((event: MessageEvent) => void) | null = null
       postMessage = vi.fn()
       start = vi.fn()
       close = vi.fn()
+
+      emit(data: unknown) {
+        this.onmessage?.({ data } as MessageEvent)
+      }
     }
 
     let channels: { port1: TestPort; port2: TestPort }[] = []
@@ -1145,20 +1150,29 @@ describe("RoomContent — Turnstile widget lifecycle", () => {
       }
     }
 
+    const localParticipant = {
+      peerId: "local-peer",
+      name: "Alice",
+      kind: "human",
+      room: "test-room",
+      muteState: false,
+    }
+
+    const remoteScreenShare = {
+      peerId: "publisher-a",
+      name: "Bob",
+      kind: "human",
+      room: "test-room",
+      screenShareEnabled: true,
+      screenShareStream: {} as MediaStream,
+    }
+
     function renderAppRoom(overrides: Record<string, unknown> = {}) {
       mockUseSfuChatRoom.mockReturnValue({
         ...baseHookReturn,
         connectionStatus: "connected",
         roomAppsEnabled: true,
-        participants: [
-          {
-            peerId: "local-peer",
-            name: "Alice",
-            kind: "human",
-            room: "test-room",
-            muteState: false,
-          },
-        ],
+        participants: [localParticipant],
         ...overrides,
       })
       return render(
@@ -1174,33 +1188,60 @@ describe("RoomContent — Turnstile widget lifecycle", () => {
       return frameWindow
     }
 
+    /** Answers the bootstrap so the host forwards live App messages. */
+    function completeHandshake(
+      frameWindow: { postMessage: ReturnType<typeof vi.fn> },
+      appId: string,
+      port: TestPort
+    ) {
+      const bootstrap = frameWindow.postMessage.mock.calls[0][0]
+      act(() => {
+        port.emit({
+          type: "ready",
+          appInstanceId: roomAppInstanceId("test-room", appId),
+          handshakeToken: bootstrap.handshakeToken,
+        })
+      })
+    }
+
+    const slot = (appId: string) => screen.getByTestId(`room-app-slot-${appId}`)
+    const slotHidden = (appId: string) =>
+      slot(appId).className.includes("hidden")
+    const slotIframe = (appId: string) =>
+      within(slot(appId)).getByTestId("room-app-iframe") as HTMLIFrameElement
+    const slotHost = (appId: string) =>
+      within(slot(appId)).getByTestId("room-app-host")
+
     beforeEach(() => {
       channels = []
       vi.stubGlobal("MessageChannel", TestMessageChannel)
     })
 
-    it("mounts the active Room App on the visual Stage while the Room conversation stays", async () => {
+    it("mounts a first launch on the visual Stage while the Room conversation stays", async () => {
       renderAppRoom()
 
-      // Conversation scope starts on Room; the Stage starts on participants.
+      // Nothing is resident before the first launch, and the Stage idles on
+      // participants.
+      expect(screen.queryByTestId("room-app-iframe")).toBeNull()
+      expect(screen.getByTestId("room-stage-participants")).toBeInTheDocument()
       expect(screen.getByTestId("interaction-tab-room")).toHaveAttribute(
         "aria-selected",
         "true"
       )
-      expect(screen.getByTestId("room-stage-participants")).toBeInTheDocument()
 
       fireEvent.click(screen.getByTestId("stage-app-shared-canvas"))
 
       const host = await screen.findByTestId("room-app-host")
       const stage = screen.getByTestId("room-stage")
       expect(stage.contains(host)).toBe(true)
-      // The App is Stage content, never conversation content: the same
-      // structure keeps it out of the stacked mobile conversation pane.
+      // Stage content, never conversation content: the same structure keeps it
+      // out of the stacked mobile conversation pane.
       expect(host.closest(".room-participants-panel")).not.toBeNull()
       expect(host.closest(".room-chat-panel")).toBeNull()
       expect(screen.getByTestId("interaction-chat").contains(host)).toBe(false)
       // A Room App is a large visual surface, like screen share.
       expect(stage).toHaveStyle({ width: "75%" })
+      expect(screen.queryByTestId("room-stage-participants")).toBeNull()
 
       // Room conversation remains selected and rendered beside the App.
       expect(screen.getByTestId("interaction-tab-room")).toHaveAttribute(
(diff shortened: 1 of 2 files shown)


madawei2699 (commented):
CHANGES REQUESTED — one narrow lifecycle blocker remains.

The resident Stage model itself looks right, but `roomAppsEnabled` is not a stable catalog/kill-switch signal. In `useSfuChatRoom.connectMediaSession(true)`, an ordinary SFU/media reconnect intentionally does `roomAppsEnabledRef.current = false; setRoomAppsEnabled(false)` while the App DataChannels are rebuilt, then restores it after transport setup succeeds. #372 currently derives `roomApps = []` from that transient false and the catalog-prune effect clears `launchedRoomAppIds`, so every resident `RoomAppHost` still unmounts/closes its MessagePort during a normal reconnect.

That violates the new invariant: App session lifetime should survive presentation changes *and* ordinary Room transport reconnect while `RoomContent` remains mounted.

Please keep the fix narrow. One acceptable shape is to distinguish transient reconnect unavailability from a stable disable/catalog removal in `RoomContent`:

- launcher/active App availability can still follow `roomAppsEnabled`;
- when `connectionStatus === "reconnecting"` and `roomAppsEnabled` temporarily becomes false, hide the active App but retain the launched ids and keep resident hosts mounted from the validated curated catalog;
- if reconnect succeeds, restore/re-show using the exact same iframe/MessagePort;
- if Room Apps remain disabled once the Room is connected/failed, or the catalog truly loses the entry, prune/unmount as today.

No new protocol or persistence is needed. An explicit server-capability signal would also be correct, but is not required if the reconnect distinction can stay local and small.

Please add two load-bearing regressions:
1. launch Canvas → `roomAppsEnabled:false + connectionStatus:reconnecting` → same iframe/port stays resident (hidden is fine) → `roomAppsEnabled:true + connected` → exact same iframe/port reused;
2. stable connected disable/catalog removal still unmounts and closes exactly once.

Everything else in this head — Canvas↔Arena residency, Hide semantics, hidden App message delivery, inert/aria-hidden, Room/Task independence, bounded resident set, and final cleanup — looks consistent with #371.

madawei2699 (commented):
APPROVE — the reconnect lifecycle blocker from the previous review is closed on head `3566f9659a7439637e74f4460666e18eaf6dffbc`.

The implementation now correctly separates the stable curated App catalog from transient transport readiness: an ordinary `reconnecting + roomAppsEnabled=false` hides the active App without pruning resident hosts, and recovery reuses the same iframe/MessagePort/session. A stable `connected|failed + roomAppsEnabled=false` still tears residents down exactly once. The two new regressions directly cover both sides of that boundary.

The original #371 invariants remain intact: Stage navigation is visibility-only, Close=Hide, hidden hosts remain non-interactive but live, resident App count stays bounded, Room/Task conversation state remains orthogonal, and Room unmount/stable disable remain the teardown boundaries. No protocol, persistence, heartbeat, PeerConnection, Lab, or domain-state expansion.

Exact-head CI is green. `cf-sfu` has since advanced via #373 only in `UserCard`/CSS/test files, with no overlap with this PR's `RoomContent` files; GitHub reports the PR mergeable/clean.

No remaining merge blocker.

PR #351, opened 2026-09-11
[codex] fix(task): stabilize Agent Task interaction authoring and artifacts

madawei2699:
Closes #348
Closes #349
Refs #350

## Summary

- Make Live View authoring safer by accepting a compact Agent draft and deriving trusted task/authority metadata at the Room boundary.
- Add deterministic local `live-view publish` preflight errors for malformed components, actions, data bindings, bounds, and payload size.
- Keep the local Runtime validator aligned with the Room's JavaScript UTF-16 text-length semantics, including CJK and emoji.
- Add task-scoped attachment metadata through the existing bounded attachment store, with participant-scoped Agent delivery and Task-only UI projection.
- Keep hidden Task attachments in the shared Agent cursor sequence as invisible placeholders, enforce Task participation for direct attachment reads, and prevent cross-Task attachment references.
- Keep Agent card voice/task controls compact inside the existing fixed card footprint.
- Update Agent prompt/bootstrap documentation and generated `llms-full.txt` for the new attachment and Live View contracts.

## Live View analytics

The same bounded rollout includes:

- `LiveViewPublished`: Room-authoritative; `phase=first|replacement`; coarse Room dimensions only.
- `LiveViewVisible`: browser-side; deduped by Task + surface.
- `LiveViewInteracted`: first meaningful local Button/Input interaction only; deduped by Task + surface.

Analytics sends no Task IDs, participant IDs/names, input values, labels/component content, or per-click/per-keystroke telemetry.

## Why

Live View publication previously required Agents to provide server-trusted identity fields, and malformed authoring errors were often discovered only after a request crossed the Runtime boundary. Attachments also had no task correlation, so Task artifacts could not be kept within the intended Task projection. These changes keep authority and task identity at the Room boundary, provide actionable local validation, and reuse the existing bounded attachment path without introducing new storage or lifecycle infrastructure.

## Validation

- `cd app && yarn test` — 81 test files / 729 tests passed
- `yarn type-check`
- `yarn eslint src --no-fix`
- `yarn prettier --check .`
- `yarn docs:check`
- `yarn cf-build`
- `cd agent && go test ./...`
- `go vet ./...`
- `go build ./cmd/free4chat-agent`
- `bash scripts/test-install-agent.sh` — 19 passed
- `bash scripts/test-bootstrap-contract.sh`
- `git diff --check`

[...]

PR #373, opened 2026-09-12
fix(room): compact Agent cards become presence cards (identity + one bounded action row)

### LINKED ISSUE #348: fix(room): keep compact Agent controls inside participant card
## Context

Observed during the 2026-09-11 production #316 Live View / screen-share validation.

When the Room Stage is occupied by a screen share (and likewise when switching to Live View), the bottom participant strip uses compact cards. A compact Agent card that exposes both `VOICE` and `TASK` controls can overflow vertically: the `TASK` pill extends below the card boundary instead of remaining contained like the Human cards.

This is a narrow presentation bug, not a #316 Live View correctness/media blocker.

## Current evidence

Production room: `baseline-316-20260911`

Observed shape:

```text
compact Agent card
├ avatar
├ Agent badge
├ VOICE
└ TASK   ← protrudes below the card shell
```

The same production run otherwise proved screen share + Task Live View coexistence successfully.

Current implementation is in `app/src/components/UserCard.tsx`: compact Agent cards stack the Agent badge, optional activity, optional VOICE button and TASK button inside the compact shell.

## Desired behavior

- compact Human and Agent cards keep a stable aligned outer height in the participant strip;
- Agent-specific `Agent` / `VOICE` / `TASK` controls stay entirely inside the card boundary;
- no clipping or overlap with the Stage/participant-strip boundary;
- preserve current actions and accessibility;
- do not enlarge the entire participant strip disproportionately just for Agents.

## Scope

Keep this as a small layout fix. Do not redesign participant cards or change Voice/Task semantics.

Add a focused regression for a compact non-self Agent with both `voiceAvailable` and `onStartTask` enabled.

madawei2699:
Closes #348

## Root cause

Compact participant cards are fixed `8rem` shells with an `84px` width (`RoomContent` passes `w-[84px]`), so the card content box is only 66px.

The compact Agent card stacked **four** visible blocks there: avatar / name / `🤖 Agent` badge / activity + VOICE + TASK. Two independent overflows followed:

1. the blocks themselves already exceeded the shell height;
2. the shared `.participant-card__voice-button { min-width: 3.8rem }` (60.8px) is almost the whole 66px content box, so the `flex-wrap` action row from #351 could never fit VOICE + TASK on one line.

Measured on the base commit, compact Agent with `kind=agent` + `voiceAvailable` + `onStartTask` + `activity`, at the production 84px width:

```
cardHeight       128
overflowY         20
row height        61   (three stacked lines)
row bottom       -19.25 below the card shell   ← the reported TASK protrusion
child tops        106.25 / 123.75 / 151.75      ← stacked, not one row
```

## Change

Reduce the compact Agent hierarchy (no new wrapping, no taller strip, no hidden capability):

```
┌──────────────┐
│   (avatar)   │   activity = small avatar indicator (title + aria-label)
│  🤖 pi-mac…  │   identity rides the truncated name line
│   🔊    ◇    │   ONE bounded h-6, flex-nowrap row of 1.5rem icon toggles
└──────────────┘
```

- `app/src/components/UserCard.tsx` — compact-only: kind glyph inline on the name line, activity as an avatar badge, Voice/Task as icon toggles in a single `h-6 flex-nowrap` row keeping their existing `aria-label` / `title` / click semantics. Full-size card untouched.
- `app/src/styles/tailwind.css` — `.participant-card__voice-button--compact` / `.participant-card__task-button--compact`: fixed 1.5rem targets that drop the shared 3.8rem voice-button min-width which forced the wrap. Base and full-size rules unchanged.
- `app/src/components/UserCard.compactAgent.test.tsx` — focused regression: one bounded non-wrapping row, VOICE and TASK share exactly one parent, kind/activity add no vertical control row, no `.flex-wrap` anywhere in the compact card, and no more stacked blocks than a compact Human card. jsdom cannot measure pixels, so the structural invariant is pinned instead.

No RoomContent state/lifecycle, screen-share, Voice transport, Task semantics, Agent Runtime, SFU or participant-protocol change.

## Verification

Same measurement after the change (Chromium, production 84px width):

```
             shellW  shellH  overflowY  overflowX  blocks  row height  row top shared
human          84      128       0          0        3        -            -
agent          84      128       0          0        3       24          yes
agent(voice●)  84      128       0          0        3       24          yes
```

- `yarn prettier --check .` ✅
- `yarn lint --max-warnings 0` ✅
- `yarn type-check` ✅
- `yarn test` ✅ 90 files / 831 tests
- `yarn cf-build` ✅ OpenNext build complete (the 36 `Failed to copy` lines for ESM markdown packages are pre-existing upstream noise — the same 36 appear in the successful cf-sfu `55bd262` deploy log)
- `git diff --check` ✅

Acceptance: same outer height for Human/Agent compact cards (1), every control inside the shell for the fully loaded Agent (2, 3), no horizontal overflow (4), Voice/Task clickable + accessible (5, 6), Agent identity visible (7), activity available non-disruptively (8), strip layout unchanged (9), full card unchanged (10).

### DIFF
--- a/app/src/components/UserCard.compactAgent.test.tsx
+++ b/app/src/components/UserCard.compactAgent.test.tsx
@@ -1,0 +1,150 @@
+import { cleanup, fireEvent, render } from "@testing-library/react"
+import { afterEach, describe, expect, it, vi } from "vitest"
+
+import UserCard from "./UserCard"
+
+/**
+ * #348: compact participant cards are presence cards, not capability
+ * dashboards. A compact Agent card must keep the same fixed footprint as a
+ * compact Human card, so the old stacked Agent badge / activity / VOICE /
+ * TASK rows are gone: identity (kind glyph) rides the name line, activity is
+ * a restrained avatar indicator, and Voice + Task share ONE bounded,
+ * non-wrapping action row.
+ *
+ * jsdom cannot measure pixels, so these tests pin the structural invariant
+ * that makes overflow impossible: a single fixed-height row with nowrap, and
+ * no more stacked blocks than a compact Human card has.
+ */
+
+afterEach(cleanup)
+
+function agentCard(overrides: Record<string, unknown> = {}) {
+  return {
+    name: "Pi",
+    kind: "agent" as const,
+    room: "room",
+    peerId: "agent-pi",
+    compact: true,
+    voiceAvailable: true,
+    voiceEnabled: false,
+    onToggleAgentVoice: vi.fn(),
+    onStartTask: vi.fn(),
+    activity: "using_tools" as const,
+    ...overrides,
+  }
+}
+
+function compactShell(): HTMLElement {
+  const shell = document.querySelector(".participant-card-shell--compact")
+  expect(shell).toBeTruthy()
+  return shell as HTMLElement
+}
+
+/** Visible stacked blocks of the compact card (audio is layout-invisible). */
+function stackedBlocks(shell: HTMLElement): Element[] {
+  const card = shell.firstElementChild as HTMLElement
+  return Array.from(card.children).filter((el) => el.tagName !== "AUDIO")
+}
+
+describe("compact Agent card density (#348)", () => {
+  it("keeps a fully loaded Agent inside one bounded, non-wrapping action row", () => {
+    const { getByTestId, getByRole } = render(<UserCard {...agentCard()} />)
+
+    const row = getByTestId("compact-agent-controls")
+    const voice = getByRole("button", { name: "Enable voice for Pi" })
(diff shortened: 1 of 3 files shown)


madawei2699 (commented):
APPROVE — the production overflow root cause is addressed rather than papered over. Compact Agent cards now match the Human card information depth (avatar + identity + one fixed non-wrapping action row), the shared 3.8rem Voice min-width is explicitly removed only for compact controls, activity no longer consumes a vertical row, and Voice/Task semantics/accessibility remain intact. Full-size cards and Room/Stage behavior are untouched. Exact-head CI is green. No merge blocker found.

PR #383, opened 2026-09-12
fix(room-app): hide Room UI during fullscreen focus

madawei2699:
The fullscreen Room App host is fixed inside the Stage subtree, whose `.room-participants-panel` creates an isolated stacking context. The sibling Room chat can therefore paint over the host even though the host is fixed to the viewport.

RoomContent now treats fullscreen as a Room layout state: surrounding Room chrome is hidden and inert while the resident App stays in the same Stage subtree. Focus follows the resident host through transient reconnect availability changes. Exiting or closing focus restores the existing layout and split ratio without replacing the host, iframe, or MessagePort.

Validation:
- `yarn test` — 91 files, 864 tests passed.
- `yarn type-check`, `yarn eslint src`, `yarn docs:check`, and Prettier check passed.
- `NEXT_PUBLIC_TURNSTILE_DISABLED=1 yarn cf-build` completed.
- Real Chrome smoke against the production Whiteboard iframe and a local production Worker/RoomSession harness passed at desktop 1280×900 and mobile 390×844. The host covered the viewport, chat/composer were hidden and inert, bottom hit-testing landed on the App, mobile touch input reached the Whiteboard, and host/iframe/ready MessagePort identity survived enter/exit.

### DIFF
--- a/app/src/components/RoomContent.test.tsx
+++ b/app/src/components/RoomContent.test.tsx
@@ -1379,7 +1379,22 @@ describe("RoomContent — Turnstile widget lifecycle", () => {
       const iframe = slotIframe("shared-canvas")
       loadAppIframe(iframe)
       const host = slotHost("shared-canvas")
+      const port = channels[0].port1
+      const chatPanel = screen
+        .getByTestId("interaction-chat")
+        .closest(".room-chat-panel")
+      const composer = screen.getByPlaceholderText(
+        "Message the room or @ an Agent…"
+      )
+      const stage = screen.getByTestId("room-stage")
+      const roomShell = screen.getByRole("main")
+      const roomHeader = screen
+        .getByTestId("room-header-identity")
+        .closest("header")
       expect(host).toHaveAttribute("data-layout", "stage")
+      expect(chatPanel).toBeVisible()
+      expect(composer).toBeVisible()
+      expect(stage).toHaveStyle({ width: "75%" })
       expect(
         within(host).getByRole("button", { name: "Fullscreen" })
       ).toHaveAttribute("aria-pressed", "false")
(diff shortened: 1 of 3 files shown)


madawei2699 (commented):
CHANGES REQUESTED on exact head `2991b10b6ddcced558fef8c5cdabf9c90022c796`.

The stacking-context fix itself is directionally correct: focus mode now removes the sibling Room UI without remounting the resident host/iframe/MessagePort, and the added identity/restoration tests cover the production regression well.

One merge blocker remains: the new focus branch hides and inerts the existing `connectionStatus === "reconnecting"` overlay. During that same reconnect, the Room App DataChannels are torn down/recreated, so the focused iframe can remain fully interactive while App sends are temporarily unavailable. For Whiteboard this can allow local edits during a period when transport is not usable, with no visible warning/guard.

Please preserve fullscreen state and iframe identity across reconnect, but keep a reconnect guard ABOVE the focused App (or equivalently make the focused iframe non-interactive and visibly disconnected until transport readiness returns). Do not collapse/exit fullscreen and do not remount the host.

Add regression coverage for: focused App -> reconnecting => focus remains active, same iframe/MessagePort remain resident, Room chat stays hidden, reconnect indicator/guard is visible and intercepts interaction; reconnect completes => guard disappears and the same App becomes interactive again.

After that narrow fix I do not currently see another blocker.

GitHub does not allow this authenticated identity to submit a formal REQUEST_CHANGES review on its own PR, so this is recorded as a COMMENT with an explicit blocking verdict.

madawei2699 (commented):
APPROVE on exact head `02c07c148d2a3ec05d4d0c703e5e5d6a4f3ebb42`.

The previous reconnect blocker is fixed correctly. Fullscreen remains a Room layout state, the resident host/iframe/MessagePort stay in place, Room chat/composer remain hidden, and reconnect now restores the visible guard above the focused App while making the App slot inert until transport returns. The regression test covers the same host/iframe/port across reconnect and recovery, and the reported real Chrome smoke covers desktop/mobile hit-testing.

I do not see another merge blocker in this diff. Current PR state is open, mergeable, non-draft; App Tests and Lint & Type Check pass on this exact head and Build & Deploy is skipped by PR policy.

GitHub does not allow this authenticated identity to formally APPROVE its own PR, so this COMMENT records the explicit approval verdict.

Issue #375, opened 2026-09-12
product(room-app): keep the core host boundary thin while discovery ownership moves to Lab


madawei2699:
## Status

**ACTIVE TRANSITIONAL UMBRELLA — Room App architecture is complete; product/discovery ownership is moving out of core.**

Canonical Lab control-plane proposal: https://github.com/i365dev/free4chat-extension-lab/issues/73  
Measurement guardrail: #346  
Core product boundary: #51

## Updated ownership decision

Free4Chat core should own the **safe temporary Room boundary**, not the continuing Room App portfolio/content lifecycle.

Today the public core repository still contains two kinds of App-specific product state that should become Lab-owned:

```text
1. hard-coded production Room App catalog entries
2. /apps discovery / SEO pages and their sitemap lifecycle
```

That creates recurring core PRs for work that is really acquisition/portfolio work: adding an App, promoting it, changing SEO copy, retiring it, preserving an old landing, or redirecting a stale slug.

The target in Lab #73 is:

```text
free4chat-extension-lab
→ canonical curated App registry
→ App runtimes
→ free4.chat/apps discovery/SEO content
→ App sitemap
→ retirement redirects / 410 decisions
→ Lab CI deployment

free4chat core
→ protocol / sandbox / MessagePort host
→ participant projection + DataChannel transport
→ fixed trusted Room App origin policy
→ generic launch seam
→ validated consumption of the Lab-owned catalog
```

After that one-time migration, ordinary App add/remove/promote/retire work should require **zero Free4Chat-core source changes**.

## Core security boundary must remain core-owned

Moving the catalog does **not** mean arbitrary remote iframe registration.

Core should continue to pin the trusted production origin:

```text
https://room-apps.free4.chat
```

A Lab-owned catalog may provide bounded validated IDs/labels/paths/status, but must not remotely choose arbitrary origins or weaken sandbox/protocol limits.

The one-time consumer seam should use strict schema/version/path/id validation and fail closed on malformed metadata.

Development fixtures may remain explicitly core-local if that is simpler.

## Generic App launch seam

The App acquisition site should not duplicate core room-name/localStorage/bootstrap details.

Add one stable generic core route/flow, conceptually:

```text
/open-app?app=<id>
```

that validates a current App ID, generates/opens a normal temporary Room, and enters it with `app=<id>` selected.

Then every Lab-managed landing page can link to one stable core capability rather than carrying App-specific launch code.

## Discovery / SEO routing

[...]

Issue #73, opened 2026-08-22
explore: social room games with Human and Agent participants


madawei2699:
## Idea

Free4Chat rooms are not only for utilitarian communication. People also open temporary rooms for social and entertainment reasons.

Once Agents are first-class room participants, a natural product direction is:

> **Humans and Agents can play lightweight room-native games together.**

The interesting part is not adding a separate game platform. It is using the existing temporary room, participant model, text/voice interaction, targeting, and user-owned Agents to create social experiences.

This is an exploration issue, not an implementation commitment.

## Product hypothesis

An Agent can participate in a room in several social roles:

- player;
- host;
- referee/judge;
- storyteller;
- quiz master;
- scorekeeper;
- improv partner;
- NPC / role-play participant.

This may make an otherwise empty/small room more entertaining and give groups something to do together without requiring a separate app/account/game install.

The same principle from #72 should hold:

> Free4Chat provides the room and shared interaction semantics. The Agent's intelligence belongs to the participant/user, not to a centrally hosted Free4Chat AI service.

## Start with games that fit the room we already have

The first candidates should work with ordinary room text/actions and should not require secret server-side state or complex real-time physics.

Good exploration candidates:

### 20 Questions / Guessing games

- one participant thinks of an object/person/concept;
- others ask yes/no questions;
- an Agent may be the guesser, host, or answerer;
- easy to run over text and later voice.

### Trivia / quiz room

- Agent acts as quiz master;
- Humans and/or other Agents answer;
- simple scorekeeping;
- categories/difficulty can be chosen conversationally.

### Story / improv chain

- Humans and Agents alternate adding to a story;
- Agent can enforce a theme/rule or introduce prompts;
- especially natural once voice participation exists.

### Word and language games

Examples:

- word association;
- forbidden-word challenge;
- riddles;
- collaborative poem/story;
- vocabulary/language practice.

These need little or no special UI initially.

### Lightweight judge games

Agent can act as a neutral judge for open-information prompts, for example:

- funniest answer;
- best explanation;
- closest estimate;
- creative challenge.

The result should be framed as entertainment, not authoritative scoring.

## More complex future possibilities

Potentially interesting but should not be first implementation:

- social deduction / Werewolf-like games;
- hidden-role games;
- Codenames-like private clues;
- poker/card games;
- board-game state;
- games requiring private participant channels or cryptographically fair hidden information.

[...]

PR #393, opened 2026-09-14
feat(room-app): load curated catalog from Lab control plane

madawei2699:
Refs [Extension Lab control-plane issue #73](https://github.com/i365dev/free4chat-extension-lab/issues/73)
Refs #375

Base: `cf-sfu` at `68e8aa7ef4f94bdb6b14ed76931185f76c4b0b71`.

Moves the curated Room App catalog authority out of Core. Core fetches a bounded v1 catalog only from the fixed trusted Lab origin, validates exact IDs/paths and size, uses the Lab catalog for authoritative unicast validation, and fails closed before a valid catalog has been accepted. Room App iframe/MessagePort/transport protocol is unchanged.

Core retains the existing `/apps` and `/apps/whiteboard` pages as temporary compatibility fallbacks and keeps both URLs in its sitemap until the Lab `/apps*` route is live. `robots.txt` advertises only the available Core sitemap; the Lab `apps-sitemap.xml` is not advertised before its route exists. The generic `/open-app` launcher and existing Room host boundary remain. Test fixtures use synthetic catalog data rather than duplicating the product portfolio.

Staged rollout: first merge/deploy Lab #78 with only the `room-apps.free4.chat/_catalog.json` route; it deliberately does not claim `www.free4.chat/apps*`. Once the catalog Worker exists, merge/deploy this Core PR and require a successful production deployment; verify `/open-app` works before continuing. Only then open a separate, small Lab cutover PR to add `www.free4.chat/apps*`. This preserves Service Binding target-first ordering and prevents discovery CTAs from going live before the launcher. No Cloudflare resources were changed in this PR. #375 remains transitional and is intentionally not closed here.

Validation:
- App: 92 test files / 900 tests passed
- type-check, ESLint, Prettier, docs:check: passed
- `yarn cf-build`: Next production compile and static generation passed, including `/apps` and `/apps/whiteboard`; local OpenNext bundling then failed because esbuild cannot load the macOS sharp native `.node` module. The PR Build & Deploy job is skipped by the current workflow, so a full Cloudflare bundle is not yet verified.
- `git diff --check`: passed
- GitHub Actions for this follow-up head: pending after push; Build & Deploy is expected to skip on PR

## Review follow-up: bounded catalog refresh and staged cutover

The browser now refreshes the fixed Lab catalog on the same 60-second cadence as Room authority revalidation (with a 59-second browser cache TTL), so removed Apps disappear from a long-lived Room instead of remaining launchable in stale UI. Development uses that same fixed-origin catalog with exact localhost CORS origins; analytics remain production-only. Catalog responses are streamed and cancelled as soon as the 16 KiB byte cap is exceeded.

The initial Lab deployment intentionally owns only the catalog route. `/apps*` stays on its current owner until this Core PR is deployed and `/open-app` is verified; a separate Lab cutover change enables discovery afterward.

## Review follow-up: compatibility fallback and deep-link recovery

The original Core `/apps` and `/apps/whiteboard` pages and their sitemap entries remain live as temporary fallbacks through the cutover. `robots.txt` does not advertise the Lab `apps-sitemap.xml` until that Lab route is actually deployed. Once `/apps*` is routed to Lab, those specific Lab routes supersede the dormant Core pages; cleanup is deferred until after cutover verification.

The direct-launch attempted marker is now set only after the loaded catalog confirms the requested App. An initially empty/unavailable catalog therefore does not consume the deep link; a later refresh can satisfy it without a separate retry mechanism.


madawei2699 (commented):
BLOCKING REVIEW — CHANGES REQUESTED

I checked current head `886ae301f33b186ede7a90c9938bc54579a96e08` and the actual catalog/RoomContent/RoomSession/open-app changes. The architecture is directionally correct, but I would not merge this head yet.

The automated Codex review already found three real issues on this exact head, and I agree they should be addressed rather than waved away:

1. Development currently loses Room Apps completely. `RoomContent` skips remote catalog loading outside production while the old local catalog has been removed, so documented local `yarn dev` has an empty App surface. Preserve an explicit development-only seam without reintroducing a production hard-coded portfolio (for example an env/config-driven dev catalog/origin or another small explicit dev mechanism).

2. Browser and Room authority lifecycle can diverge in long-lived Rooms. The browser snapshots the catalog only at mount, while `RoomSession` revalidates unicast against the 60-second catalog loader. If an App is disabled, the UI can keep it mounted/offered while unicast starts returning `app_unavailable`. Pick and document one lifecycle semantic and make both sides consistent. A simple per-Room/session snapshot is acceptable if retirement should affect new Rooms only; periodic client refresh is also acceptable if immediate disable is intentional. Do not build a complex invalidation system.

3. The 16 KiB catalog bound is checked only after `response.text()` has buffered the body. If the bound is meant to protect the trust seam, read/cancel bounded bytes rather than buffering an arbitrarily large body first.

There is also a rollout dependency with Lab #78: Cloudflare service bindings require the target Worker to exist before this Worker can deploy, so Lab's control-plane Worker must exist first. But Lab #78 currently also takes over `www.free4.chat/apps*` and its CTA points to this PR's new `/open-app`; that creates a public broken-CTA window if Lab is deployed first and this Core deploy is delayed/fails. Please coordinate with #78's staged cutover so the catalog/control-plane service can exist before Core while `/apps*` is not switched until `/open-app` is live.

Given this PR changes the Worker service binding and local `cf-build` did not complete OpenNext bundling, I would also want either a Linux/CI production bundle/dry-run or a rollout plan where failure of this deploy cannot strand `/apps` on a new page whose launcher is unavailable. No need to add enterprise release machinery; just make the one-time migration failure-safe.

The fixed trusted origin, strict remote schema/path validation, no baked-in production portfolio, generic launcher, and removal of App-specific SEO ownership from Core are otherwise the right direction.

madawei2699 (commented):
BLOCKING REVIEW — CHANGES REQUESTED

Re-reviewed current head `158eaabe396b6714285cc388e3d127c91c31aece`.

The prior technical blockers are fixed: development now loads the Lab catalog, long-lived Rooms refresh and retire removed Apps, and the catalog body is streamed/cancelled at the 16 KiB bound. The new tests cover those paths and current CI is green.

One rollout blocker remains in the staged plan itself. This PR still deletes Core's existing `/apps` and `/apps/whiteboard` pages and removes them from the Core sitemap. Lab #78 now intentionally deploys only `room-apps.free4.chat/_catalog.json`; its `www.free4.chat/apps*` route is deferred until after this Core PR is deployed and `/open-app` is verified. Therefore, immediately after this Core deploy and before the later Lab cutover PR, `/apps` and `/apps/whiteboard` will fall through to Core with those pages removed, creating a public discovery/SEO outage. `robots.txt` will also advertise `https://www.free4.chat/apps-sitemap.xml` before the Lab Worker owns that `www.free4.chat` route.

Please keep the existing Core `/apps` and `/apps/whiteboard` pages as temporary compatibility fallbacks through this PR (they become unreachable once the more-specific Lab route is cut over), and keep the current App sitemap entries/robots behavior coherent until that cutover; then remove the dormant compatibility pages/sitemap ownership in the small Lab/Core cleanup after the route is live. An equivalent failure-safe sequencing is fine, but there should be no window where the existing acquisition URLs or advertised sitemap return 404.

Small resilience follow-up while touching this area: `initialRoomAppLaunchAttemptedRef` is currently set before confirming that the requested App exists in the loaded catalog. A transient first catalog failure returns an empty fail-closed catalog, marks the direct-launch attempt consumed, and a later 60-second recovery will show Apps but will never auto-open the originally requested `?app=`. Move the attempted marker to the successful membership/launch path (or otherwise allow a later valid catalog to satisfy the original deep link). This is a narrow fix; no retry framework is needed.

With those two migration/deep-link details fixed, the Core design looks ready.

madawei2699 (commented):
APPROVE

Reviewed current head `53670846298c69ab8829f67e3c582ceedcc17ab5` against the two remaining blockers and the earlier control-plane concerns. I do not see a remaining code blocker.

The compatibility window is now safe: Core keeps `/apps` and `/apps/whiteboard` plus their current sitemap entries until the later Lab `/apps*` cutover, and it does not advertise `apps-sitemap.xml` before that Lab route exists. This removes the 404/SEO gap between Core deployment and discovery cutover without making the compatibility pages canonical again.

The direct-launch retry semantics are also corrected: an initially empty/unavailable Lab catalog no longer consumes `?app=<id>`; the attempted marker is set only after catalog membership is confirmed, and the new test covers recovery on the next catalog refresh.

The previous blockers remain fixed on this head as well: development loads the fixed Lab catalog, long-lived Rooms refresh/retire catalog entries consistently with Room authority, and catalog bodies are stream-bounded/cancelled above 16 KiB. Current GitHub App Tests and Lint/Type/Docs checks are green. The PR is mergeable.

The local macOS `sharp` failure means the full OpenNext bundle still has not been proven on this PR head, but given the staged rollout it is no longer a merge blocker: Lab #78 must deploy the catalog-only Worker first, existing `/apps*` remains intact if Core deployment fails, and the later Lab discovery cutover is gated on successful production `/open-app` verification.

Merge/deploy order remains important: deploy Lab #78 catalog-only first; then merge/deploy this PR and smoke `/open-app` plus an existing App; only afterward cut over Lab `/apps*`. Do not close #375/#73/#24 until the corresponding production cutover/cleanup is verified.

PR #396, opened 2026-09-14
fix(room-app): preserve host transport across catalog refresh

madawei2699:
## What

Fixes a production Room App sync regression introduced/exposed by #393 (`feat(room-app): load curated catalog from Lab control plane`, merge `a200749`).

`RoomAppHost` now owns its iframe `MessagePort` bridge for the mounted App instance lifetime instead of tying teardown to `RoomAppDefinition` object identity. One file of production code changes (`app/src/components/RoomAppHost.tsx`); the rest of the diff is regression tests.

## Why

#393 added a 60s browser catalog refresh:

```ts
setInterval(refreshCatalog, ROOM_APP_CATALOG_REFRESH_INTERVAL_MS)
```

Every refresh re-parses the Lab catalog (`parseRoomAppCatalog(JSON.parse(body))`) and installs it through `setLoadedRoomAppCatalog(catalog)`. Even when nothing changed, React receives **brand-new `RoomAppDefinition` objects**.

`RoomAppHost` used to own transport teardown from the whole `app` object:

```ts
useEffect(() => {
  const pendingUnicastRequests = pendingUnicastRequestsRef.current
  if (!validateRoomAppDefinition(app) || !isRoomAppAllowlisted(app)) {
    setFailed(true)
    return
  }
  return () => {
    readyRef.current = false
    portRef.current?.close()
    portRef.current = null
    pendingUnicastRequests.clear()
  }
}, [app])
```

So every 60s, for every resident App:

```text
catalog refresh
→ new RoomAppDefinition object identity
→ RoomAppHost effect cleanup
→ portRef.current.close(); readyRef.current = false
→ iframe is NOT reloaded (same src, same key)
→ sendBootstrap() only runs from onLoad, so no new MessageChannel
→ iframe ↔ host bridge permanently dead
```

Underlying Room transport (text, voice, DataChannel, even the App DataChannel lane) stays healthy, which is exactly the reported symptom: ordinary Room works, but every Room App silently stops syncing (Whiteboard sometimes, Timer sometimes, everything eventually), and once broken it never recovers.

Confirmed at HEAD `6ae60af` by inspection, by unit/integration regression tests that fail on the old implementation, and by a real two-page browser reproduction (see below: bridge closed at ~61s after launch, all App sync dead afterwards).

## How

`app/src/components/RoomAppHost.tsx` only:

- **Validation/allowlist becomes a metadata-only reaction.** It still drives the unavailable state, but it no longer owns the bridge.
- **Transport teardown is bound to the real App instance lifetime** — `appInstanceId` / real unmount — via one `retireTransport()` helper, never to `RoomAppDefinition` object identity. A catalog refresh that still resolves the App id keeps the resident iframe and its `MessagePort`.
- **The failure transition retires the bridge too**, because that state removes the iframe; `retireTransport()` also clears the `ready` UI, so a retired bridge can no longer display `ready` (no heartbeat added).
- **Real iframe navigation still re-bootstraps** through the existing `onLoad → sendBootstrap()` path, which already closes the obsolete port and creates a fresh `MessageChannel`; a metadata-only change never reaches it.

Deliberately unchanged: App protocol (`protocolVersion: 1`, message shapes, rate limits), DataChannel lanes, Lab control-plane ownership, catalog schema/refresh cadence, App domain-state ownership, `RoomContent` residency rules. Nothing forces a re-bootstrap per refresh.

## Testing

New regression tests (all in `app/src/components/RoomAppHost.test.tsx` and `app/src/components/RoomContent.test.tsx`):

1. **Identical catalog refresh keeps the transport alive** — launch App, complete handshake, capture iframe + `MessagePort`, refresh with a logically identical catalog parsed into new objects, then assert the same iframe DOM node, exactly one `MessageChannel`, `port.close()` never called, no hidden re-bootstrap, `App → host` reliable send still reaches `sendRoomAppMessage`, and `host → App` remote reliable still lands in the iframe. (RoomContent-level test drives the real 60s interval + real catalog parser + real launch/handshake helper.)
2. **Metadata-only refresh does not reset transport** — same App id/path/origin with a changed label: UI label updates in the Stage picker and host chrome, iframe/port identical, no close, reliable traffic continues.
3. **Catalog removal really tears down once** — resident App removed from the refreshed catalog: `RoomAppHost` unmounts, iframe removed, `port.close()` exactly once, and a later refresh does not close it a second time.
4. **Real iframe navigation can still re-bootstrap** — a genuine second `onLoad` on the same resident iframe closes the old port exactly once, creates a new `MessageChannel` with a new handshake token, returns the status to `connecting…`, completes the new handshake, and forwards realtime traffic again.
5. **Truthful status** — an unusable definition retires the bridge and no longer reports `ready`.

Pre-fix evidence (same tests, `RoomAppHost.tsx` reverted to `6ae60af`):

```text
× keeps the handshaken bridge when a catalog refresh returns an equal new definition
    AssertionError: expected "vi.fn()" to not be called at all, but actually been called 1 times
× keeps the handshaken bridge when only App metadata changes
    AssertionError: expected "vi.fn()" to not be called at all, but actually been called 1 times
× never leaves a ready UI on a retired bridge when a definition becomes unusable
    AssertionError: expected "vi.fn()" to be called 1 times, but got 0 times
× keeps a resident App's host transport across an identical catalog refresh   (RoomContent)
× applies a metadata-only catalog refresh without resetting a resident App   (RoomContent)
```

Tests 3 and 4 pass before and after on purpose: they guard that retirement and navigation re-bootstrap are not broken by this fix.

Validation on head `ef3fd4c`:

- `yarn test`: 92 test files / 907 tests passed
- `yarn type-check`: passed
- `yarn lint --max-warnings 0`: passed
- `npx prettier --check .`: passed
- `yarn docs:check` + CI doc/discovery vitest subset: passed
- `yarn cf-build` (OpenNext/Cloudflare Worker bundle): Next production compile, 25/25 static pages, `OpenNext build complete`
- `git diff --check`: passed

## Real browser smoke (two pages, one Room, >60s)

Two isolated Chrome contexts joined one Room against the locally built Worker (`wrangler dev` on the real `.dev.vars` SFU credentials, local DO), each launched the App, and the script instrumented `window.MessageChannel` to identify the host bridge by the host message shape it carries (`type` + `appInstanceId`):

- **Pomodoro, fixed build** — initial sync `Alice → Bob`, then 75s (crossing the 60s refresh): 1 host bridge per page, `roomAppClosed = 0`, `roomAppCloseAt = []`, bridge posts increasing across the refresh, post-refresh sync **Bob → Alice** and **Alice → Bob** both converge, Room UI and WebRTC voice subscription stay connected. 15/15 checks passed.
- **Whiteboard (Excalidraw), fixed build** — 0 → 1 → (cross 77s) → 2 → 3 items, both pages identical at every step, 1 host bridge, 0 closes, traffic increasing across the refresh. 15/15 checks passed.
- **Ordinary Room health on the same fixed build** (separate check, no App involved) — human text round-trip `Alice → Bob` appears in the peer timeline, both pages hold a live remote audio subscription (`audio.srcObject` with real audio tracks from the Cloudflare Realtime SFU), no reconnect/disconnect in either console. 5/5 checks passed.
- **Same Pomodoro smoke on the pre-fix build** — reproducible production failure:

```text
[check] FAIL no host bridge closed across refresh (Alice)  roomAppCloseAt: [61232] ms
[check] FAIL no host bridge closed across refresh (Bob)    roomAppCloseAt: [60771] ms
[check] FAIL post-refresh sync Bob -> Alice   {"alice":"Pause","bob":"Resume"}
[check] FAIL host bridge still open at the end (Alice)     roomAppClosed: 1
```

i.e. the host bridge closes 60.7–61.2s after launch and every post-refresh App action stays local: exactly the reported production symptom.

## Notes

Not merged here. No Lab App runtime, catalog architecture, control-plane ownership, protocol or domain-state change was made.


madawei2699 (commented):
APPROVE — reviewed exact head `ef3fd4cf1591d9cd429baac9eb67fb93b57c38d9` against current `cf-sfu` tip `6ae60af87e655fc3bb06961aa9d4cd6f4094fedf`.

The fix is at the correct lifecycle boundary: catalog/metadata object identity no longer owns the iframe MessagePort; transport cleanup follows the mounted App instance / true retirement, while a real iframe load still owns re-bootstrap and replacement of an obsolete port. `failed` retirement now also clears the rendered ready state instead of leaving a silent dead bridge reported as ready.

The regressions exercise the important production failure rather than only object equality: a freshly parsed equivalent catalog object and a metadata-only relabel keep the same iframe and same port with bidirectional reliable traffic still working; catalog removal closes once; a genuine subsequent iframe load replaces the port and handshakes again. The RoomContent test additionally drives the real catalog refresh interval/loader path.

No protocol, DataChannel, catalog schema, Lab control-plane ownership, or App runtime/domain-state change is introduced. Exact-head App Tests and Lint/Type Check are green, and the supplied >60s Pomodoro/Whiteboard browser reproduction provides strong causal confirmation of both the old failure and the fix.

No merge blocker found.

Issue #406, opened 2026-09-16
security(realtime): restore production Turnstile and bound anonymous cost-abuse admission


madawei2699:
## Status

**READY — PR 1 should be narrow and production-focused.**

This issue records a cost-abuse review of Free4Chat's anonymous realtime admission boundary. The immediate production fix is small: restore the already-designed Turnstile protection for fresh Human SFU sessions and make production fail closed when the Turnstile secret is missing.

Do **not** turn PR 1 into a general rate-limiter rewrite, WAF project, account system, CAPTCHA redesign, or RoomSession refactor.

Current reviewed branch/head at issue creation:

```text
branch: cf-sfu
head: f8ac9bca0127f01aa84252464d5c9e38783c4cde
```

## Why this matters

Free4Chat is intentionally anonymous and cheap to join. That means the realtime entry path itself is a cost boundary:

```text
untrusted Internet
  ↓
Worker request
  ↓
Turnstile / rate limit
  ↓
Cloudflare Realtime session creation
  ↓
RoomSession Durable Object registration
  ↓
WebSocket / SFU / optional Room Apps
```

A malicious caller does not need to compromise user data to cause damage. Repeated anonymous resource creation can amplify into:

- Worker requests;
- KV rate-limit operations;
- Durable Object invocations / storage / alarms;
- Cloudflare Realtime sessions and egress;
- downstream Room/App resources after admission.

Core is also the trust root for future Room/App capabilities. Extension-level cost controls cannot protect Core if Core admission itself is easy to automate.

## Current production finding

The application already has the intended action-scoped Turnstile design from #75:

```text
new Human session
→ request fresh single-use Turnstile token
→ POST /api/sfu/session
→ server-side Siteverify
→ create Cloudflare Realtime session
→ register Human in RoomSession DO
```

Reconnects intentionally use the existing participant/session capability and do not require a new challenge.

However, the current production deploy workflow still carries the temporary dogfood bypass:

```yaml
Build:
  NEXT_PUBLIC_TURNSTILE_DISABLED: "1"

Deploy:
  --var "TURNSTILE_DISABLED:true"
```

The comments describe this as a temporary Agent Voice dogfood window, but the current `cf-sfu` production deployment still uses it.

In addition, server verification currently fails open when the secret is absent:

```ts
if (env.TURNSTILE_DISABLED === "true") return true
if (!env.TURNSTILE_SECRET_KEY) return true
```

[...]

PR #407, opened 2026-09-16
security(realtime): bound anonymous realtime cost and restore production Turnstile (#406)

madawei2699:
Refs #406 — complete cost-abuse hardening, one coherent invariant:

> **Anonymous or cheap Free4Chat admission must not be able to create unbounded realtime cost. Idle RoomSessions must remain hibernatable; room/participant/state growth and expensive ingress must be bounded.**

Audit base: `cf-sfu` @ `f8ac9bca0127f01aa84252464d5c9e38783c4cde` (re-fetched before branching; still the current head — no drift from the issue baseline). New head: `ee72ef2`.

Five commits, logically separated:

| Commit | Scope |
| --- | --- |
| `1017409` | production Turnstile restored + fail closed without the secret |
| `8c44516` | RoomSession: no duration pinning, participant/state/connection/message bounds |
| `e1271fe` | Workers Rate Limiting bindings + pre-DO probe guards + Turnstile action |
| `837688e` | docs for the combined invariant |
| `ee72ef2` | **every handle-based MCP tool guarded pre-DO + server-enforced wait cadence** |

---

## 1. Turnstile (kept from the first commit, unchanged)

* production Build no longer sets `NEXT_PUBLIC_TURNSTILE_DISABLED=1`;
* production Deploy no longer passes `--var "TURNSTILE_DISABLED:true"`;
* missing `TURNSTILE_SECRET_KEY` (with the bypass off) → **`503 turnstile_not_configured`** before any Cloudflare Realtime session or RoomSession registration, never a fake `verification_failed`;
* reconnect stays capability-authenticated: no Turnstile, no token, and provably no secret needed;
* public pages stay ungated (no global gate returns); the explicit local/test/E2E bypass remains;
* **new** — the existing widget now renders `action: "sfu-session"` (one shared constant) and Siteverify must echo it back. A token minted by another widget on the same domain-locked sitekey is rejected; an *absent* action is tolerated so pre-#406 client builds keep working. Hostname validation was deliberately **not** added: the sitekey is already domain-locked, and `www`/apex/localhost all legitimately appear depending on environment (documented in code).

## 2. `wait_for_events` no longer pins DO duration

A parked long-poll needs both an in-flight request and a `setTimeout` — exactly the two things that make a Durable Object non-hibernatable — so one Agent could chain 25 s waits and keep a Room billable at ~1 request/25 s (≈ $4.15/room-month at 128 MB).

* Normal resident delivery was already the hibernatable `/api/room/agent-events` WebSocket (`agent/internal/free4chat/resident_events.go`; the Go tests assert the official Runtime never calls public `wait_for_events`) — **no Runtime change was needed**.
* The public MCP call now returns the current event snapshot immediately, still renews the Agent lease, and reports `longPoll: "immediate" | "held" | "cooling_down"` plus `retryAfterMs` when it did not hold.
* The held path remains for compatibility but is an explicit per-environment opt-in (`MCP_LONGPOLL_ENABLED`), clamped to **5 s**, with a **15 s per-Room idle gap** after each reserved window — long enough for the runtime to hibernate, so holds can never be chained into a permanently awake Room.
* No `awakeMs` accounting system was introduced: the cause was removed, not metered. Regression probe: with the default config an `agent-wait(timeoutSeconds: 25)` leaves **zero pending timers** and zero parked waiters (`vi.getTimerCount() === 0`); with the opt-in the clamp and the gap are asserted.

## 3. Room participant bound

`MAX_ROOM_PARTICIPANTS = 32`, enforced in the shared `register` / `agent-register` admission branch for Humans and Agents alike → `409 room_full`. A duplicate id still reports `participant_exists`, and `reconnect` neither adds a participant nor consults capacity (tested at full capacity). No quotas/accounts/matchmaking. 32 was checked against every existing fixture (largest single Room in the suite: 9).

## 4. Primary Room record byte budget

`MAX_PRIMARY_ROOM_BYTES = 80 KiB` UTF-8 serialized (backend value limit is 128 KiB), measured **before** `storage.put`, so an oversized value can never reach storage.

* Bounded history is evicted oldest-first, keeping the newest message: chat stays writable and the record can never exceed the budget through legal messages (tested with 40 × 2 000 CJK characters).
* Anything the eviction order cannot free (a single oversized `actionPayload`, or non-evictable state) → deterministic **`507 room_state_budget_exceeded`** / the equivalent WebSocket `{type:"error"}` frame, with the previously persisted state untouched (asserted byte-for-byte via a post-load snapshot).
* A legacy record that predates the budget stays **readable** and self-heals on the next mutation rather than wedging every load.

## 5. Expensive admission → Workers Rate Limiting binding

New shared `allowAdmission` helper; production declares three bindings in `wrangler.jsonc` (visible in `wrangler deploy --dry-run`):

```text
env.SFU_ADMISSION_RATE_LIMITER (20 requests/60s)   /api/sfu/session, /api/sfu/agent-session
env.MCP_JOIN_RATE_LIMITER      (10 requests/60s)   MCP create_room, join_room
env.ROOM_PROBE_RATE_LIMITER    (60 requests/60s)   unauthenticated probes (room_info, /api/sfu/ws)
env.MCP_HANDLE_RATE_LIMITER    (240 requests/60s)  every participantHandle-bearing MCP tool
env.MCP_WAIT_RATE_LIMITER      (1 request/10s)     wait_for_events cadence
```

* Same budgets as before, keyed by **IP + operation class**; the KV counter is no longer touched on these paths in production (tests assert KV get/put are not called), so the per-attempt KV read+write cost amplifier is gone. The KV branch survives only as a fallback for environments without the binding (unit tests / older harnesses) and for a binding failure, so a limiter outage never becomes a service outage.
* Stated honestly: Workers Rate Limiting is **coarse, per-location, eventually consistent** — a flood damper, never an exact global quota or an identity system. Authenticated Room actions are deliberately **not** IP-limited (NAT/shared-network safety); those are bounded by capability/semantic caps instead.

## 6. **Every** ingress that resolves a Room is guarded before any DO call

`participantHandle` is base64url JSON, not a signature: `decodeHandle` proves nothing and the token is only checked *inside* RoomSession. So **any** handle-bearing tool used to resolve an attacker-supplied `roomId` into a DO invocation before any authorization — a forged handle on a rotating roomId (or a rotating tool name) was a free DO wake-up. `room_info` and `/api/sfu/ws` were guarded in `e1271fe`; `ee72ef2` closes the rest:

* one shared pre-DO guard (`guardedRoomControl`) covers **all** handle-bearing tools — `read_room_context`, `wait_for_events`, `send_text`, `update_capabilities`, `update_runtime_host`, the collab tools, `send_attachment`, the surface tools, `publish_live_view`, `leave_room` — so a newly added tool cannot forget it. Keyed by caller, never by Room identity (which the caller controls), under `MCP_HANDLE_RATE_LIMITER` (240/60s, far above a busy legitimate caller);
* unauthenticated probes (`room_info`, `/api/sfu/ws`) keep `ROOM_PROBE_RATE_LIMITER` (60/60s);
* an abusive probe gets `429 rate_limited` with **zero Durable Object invocations** (asserted with a recording DO namespace for `wait_for_events`, `send_text` and `read_surface`), while legitimate calls still reach their Room (asserted).

With that, the "rejected before any DO call" claim holds for every pre-auth Room resolution, not just two routes.

## 6b. `wait_for_events` cadence is server-enforced, not a hint

`retryAfterMs` was advisory, so a client could ignore it and tight-loop: every accepted wait also refreshed the lease and wrote the primary Room record, which is both an awake-time and a storage-write amplifier.

* **Pre-DO**: `MCP_WAIT_RATE_LIMITER` (1 per 10s per caller + participant) refuses with a shaped `{ "error": "wait_rate_limited", "retryAfterMs": 10000 }` before any Durable Object is woken.
* **In the DO**: RoomSession enforces the same floor authoritatively per participant (5s) *before* refreshing the lease or saving, so a caller that ignores the hint cannot keep a Room busy, and the refusal provably does not touch Room state (asserted).
* The resident Runtime is unaffected: it uses its own hibernatable Agent event stream, and the Go daemon tests assert it never calls public `wait_for_events`.

## 7. Connection / message amplification

Audited first, then bounded only where the code proved a gap:

* **Human browser sockets** had no cap and did not close older sockets; the only guard was participantId+token, so one capability could hold unlimited sockets that all remained able to mutate Room state. Now: **4 per participant** (older ones closed `4000 "Replaced"`) and **64 per Room** (`429 room_connection_limit`). The cap keeps older sockets alive *under* the limit, so normal reload/media-reconnect is unaffected.
* **Client messages**: no per-connection/participant limit existed (only `room-app-unicast` had a 10 msg/s per-socket budget). Now **300 per participant per 10 s** → `message_rate_limited`, dropped before touching Room state.
* **Agent event stream**: already self-replaces by tag (one usable socket per participant), so per the review guidance it was documented rather than changed.
* Lifecycle semantics are untouched: no 24 h Room lifetime, no account/session model, hibernation and empty-room expiry unchanged.

## Out of scope (unchanged, intentionally)

SQLite-backed DO migration; accounts/login/payment; a generic Auth DO; room/capability work from Lab #110; global Turnstile page gate; Turnstile for native Agents; Cloudflare Access; distributed quota infrastructure; a repo-wide RoomSession refactor; WAF management code.

---

## Validation (all run locally on `837688e`)

| Check | Result |
| --- | --- |
| `yarn test` | **96 files / 977 tests passed** (the individual commits were also verified: `8c44516` alone → 96/959 + clean `type-check`) |
| `yarn type-check` | passed |
| `yarn lint --max-warnings 0` | passed |
| `npx prettier --check .` | `All matched files use Prettier code style!` |
| `yarn docs:check` | passed (llms assets regenerated from the updated docs) |
| `yarn cf-build` | `OpenNext build complete` |
| `wrangler deploy --dry-run` | bundles `worker.ts`; **five** Rate Limit bindings present; `room_full`, `room_state_budget_exceeded`, `room_connection_limit`, `message_rate_limited`, `wait_rate_limited`, `mcp_ingress_rate_limited`, `turnstile_not_configured`, `longPoll`, `MCP_LONGPOLL_ENABLED` all present in the bundle |
| `yarn e2e:room` | **1 passed** — two real browsers join through the real local Worker + `RoomSession` DO; the spec hard-asserts `/sessions/new` was served and zero unexpected Realtime calls |
| `yarn e2e:room-app-host` | **6 passed** (desktop/phone/tablet × chromium/webkit) |
| `cd agent && go test ./...` | **not a gate here: this PR contains zero Go changes** (`git status agent/` clean). Packages that could build in this sandbox passed (`free4chat`, `harness`, `speech`, `types`, `voice`, `attachments`); the rest fail at *setup* with `operation not permitted` on `/Users/luying/go/pkg/mod/cache/download/...`, i.e. the module cache is outside the sandbox workspace — not a test failure. CI (`go-agent.yml`) will cover it. |

Local-env notes (not repo issues): this worktree needed a writable `XDG_CONFIG_HOME` for wrangler's registry and `PLAYWRIGHT_BROWSERS_PATH` for browsers; the OpenNext `Failed to copy node_modules/...` log lines are pre-existing noise (produced `.open-next` package set is identical to a previously built checkout).

## Test coverage added

* `app/src/do/roomSessionCostBounds.test.ts` (new, 11 tests): no-timer default wait + clamped/gapped opt-in hold + lease refresh; server-enforced wait cadence (refused inside the floor without touching the lease, accepted past it); participant cap (Human/Agent/duplicate/reconnect-at-capacity); byte budget (eviction keeps the Room writable; unfittable → shaped WS error and unchanged stored state; oversized control mutation → shaped 507); per-participant socket replacement; room socket cap; message budget.
* `app/src/sfu/server.test.ts`: the full Turnstile admission matrix from the first commit plus action matching/mismatch/pre-rollout tolerance, and the binding path (used instead of KV, 429 before any Realtime/DO spend, `/ws` probe guard before the DO, KV fallback only when the binding is absent/failing).
* `app/src/mcp/server.test.ts`: `room_info` probe rejected before the DO / served when legitimate; `join_room` + `create_room` throttled by binding with KV untouched; over-budget join rejected before the DO.
* `app/src/mcp/server.test.ts`: `room_info` probe rejected before the DO / served when legitimate; `join_room` + `create_room` throttled by binding with KV untouched; over-budget join rejected before the DO; **forged handles on three different tools rejected with zero DO invocations**; a legitimate handle still reaches its Room; the pre-DO wait cadence and the DO-side backstop both surface `retryAfterMs`.
* `app/src/__tests__/deploy/turnstile-production-contract.test.ts`: production deploy carries no Turnstile disable switch, no plain-var secret, and fails if any of the five rate-limit bindings is dropped.

Two existing tests were adapted to the new invariant (the wake-boundary tests now enable the explicit legacy long-poll, and the over-cap resident-envelope test drives the delivery path directly because a multi-megabyte Room is no longer reachable through stored state — its assertion is unchanged).

## ⚠️ Operational prerequisite before merge/deploy

**`TURNSTILE_SECRET_KEY` Worker-secret status: cannot confirm.** It is documented as a Cloudflare Worker secret (`npx wrangler secret put TURNSTILE_SECRET_KEY`) and is correctly absent from the repo; this environment has no Cloudflare credentials (`wrangler whoami` → not authenticated), so the production Worker's secret list could not be inspected. The new rate-limit bindings are declared in `wrangler.jsonc` and appear in the dry-run bundle, so they need no manual step.

**BEFORE MERGE/DEPLOY:** run `cd app && npx wrangler secret list` and confirm `TURNSTILE_SECRET_KEY` exists on `free4chat-realtime`. If it is missing, the first deploy makes every *fresh* Human join fail closed with `503 turnstile_not_configured` (reconnects and Agent/MCP paths unaffected) — the intended fail-closed behavior, but it must not be discovered in production.

Two deploy-time behavior changes worth knowing: with `MCP_LONGPOLL_ENABLED` unset, `wait_for_events` returns immediately, and it now enforces a 10 s server-side cadence. The official Runtime is unaffected (hibernatable event stream), but a third-party MCP client that polled harder than once per 10s will now receive `wait_rate_limited` and must honor `retryAfterMs`; set `MCP_LONGPOLL_ENABLED=true` only for a deliberate compatibility window.


madawei2699:
Scope update after the cost-abuse audit: **keep this as one coherent hardening PR; do not split the remaining #406 work into follow-up PRs unless a change truly requires a separate migration.**

The Turnstile/fail-closed work already in this PR is correct, but the audit found that the larger invariant is broader:

> **Anonymous or cheap Free4Chat admission must not be able to create unbounded realtime cost. Idle RoomSessions must remain hibernatable; room/participant/state growth and expensive ingress must be bounded.**

Please extend this same PR to finish that invariant end-to-end.

## Required additions in this PR

### 1. Keep the current Turnstile work

Retain everything already implemented:

- production `NEXT_PUBLIC_TURNSTILE_DISABLED=1` removed;
- production `TURNSTILE_DISABLED:true` removed;
- missing `TURNSTILE_SECRET_KEY` fails closed;
- reconnect remains capability-authenticated and does not re-challenge;
- public pages remain ungated;
- explicit local/test/E2E bypass remains.

Also inspect the current widget/Siteverify contract and, **if the existing client already supplies stable `action` / expected hostname data**, validate them server-side in the same PR. Do not invent a second challenge flow just to add this.

### 2. Remove the DO-duration pinning path from normal Agent event delivery

The important new finding is that `wait_for_events(timeout <= 25s)` currently keeps an HTTP request and `setTimeout` alive inside `RoomSession`, so a valid Agent can continuously long-poll and prevent hibernation with very low request volume.

Free4Chat already has the correct primitive: the hibernatable `agent-events` WebSocket.

Target invariant:

```text
normal resident Agent event delivery
→ hibernatable agent-events WebSocket
→ idle RoomSession can sleep
```

Do not leave a normal supported path where one Agent can chain 25-second HTTP waits forever and pin DO duration.

Preferred implementation:

- migrate/route the normal Runtime/Harness event wait path onto the existing `agent-events` hibernatable WebSocket;
- preserve the public MCP semantics expected by the Runtime, but the underlying normal delivery must not pin the DO;
- if a legacy HTTP `wait_for_events` fallback must remain for compatibility, make it explicitly bounded and non-default so it cannot be chained indefinitely to hold a Room awake;
- add a small emergency kill switch such as `MCP_LONGPOLL_ENABLED` only if a legacy long-poll path remains. Default production behavior should be the safe one.

Do **not** replace this with a custom `awakeMs` accounting system if the hibernatable event path can remove the cause directly.

Add a regression test/probe that demonstrates an idle resident Agent no longer leaves a non-hibernatable timer/in-flight long-poll path as its normal event wait mechanism.

### 3. Bound Room participants

`participants` currently has no explicit hard maximum.

Add one product-level bound (choose a conservative value consistent with existing Room expectations, e.g. 32/64 after inspecting tests/current UI assumptions). Enforce it at Room admission for both Human and Agent paths with a deterministic error.

Requirements:

- reconnect for an existing participant must not be blocked by the capacity check;
- duplicate/reconnect must not consume another slot;
- both Human and Agent fresh admission respect the same Room capacity invariant unless the existing architecture proves they need separate caps;
- tests cover full room, reconnect at full room, Agent/Human behavior.

Do not build quotas/accounts/matchmaking.

### 4. Bound the serialized primary Room record before storage

The legacy KV-backed DO storage value limit is finite; do not rely on `storage.put()` throwing after a large valid Room state has already been constructed.

Add an explicit serialized-byte preflight around the **primary Room record** before persistence.

Requirements:

- use UTF-8 serialized byte size, not JS string length;
- define a hard maximum safely below the backend value limit with some headroom;
- normal bounded histories remain unaffected;
- when a mutation would exceed the Room-record budget, fail deterministically / preserve the previous valid state rather than corrupting the room or letting an uncaught `put` failure become the protocol behavior;
- add a test that drives the Room through legal public inputs until near/over the cap and proves the failure is controlled.

Do not migrate this PR to SQLite-backed DO. That is explicitly out of scope.

### 5. Replace the KV `get -> put` admission limiter where it is protecting expensive creation

The current KV limiter is non-atomic, eventually consistent and itself incurs KV read/write cost on every attempt.

Use the Cloudflare Workers Rate Limiting binding for **coarse abuse throttling at expensive/pre-admission boundaries**, rather than using KV as the normal request counter.

At minimum cover the existing creation/admission paths currently using the KV limiter:

- fresh Human `/api/sfu/session`;
- `/api/sfu/agent-session` if it creates comparable realtime resources;
- MCP `create_room` / `join_room` admission.

Then inspect unauthenticated/pre-auth routes that can wake a random `RoomSession` (for example `room_info` and invalid-capability probes). Where a cheap Worker-side rate guard can reject abuse **before** resolving/calling the DO without changing legitimate semantics, include it here.

Important constraints:

- Workers Rate Limiting is coarse/per-location, **not** an exact global quota or identity system;
- do not claim atomic account-wide enforcement;
- do not mechanically IP-limit every authenticated Room action and break NAT/shared-network usage;
- prefer `IP + operation class`, and capability/room semantic bounds after admission;
- remove obsolete KV limiter reads/writes from the paths migrated to the binding so the cost amplifier is actually gone.

Do not add a generic anti-abuse service or WAF orchestration framework.

### 6. Bound authenticated message/connection amplification where current code proves a gap

Audit the existing Room WebSocket / agent-event connection lifecycle and per-connection/per-participant message handling while you are already in this PR.

If one participant capability can currently open unlimited simultaneously-active connections or send unbounded messages with no existing generation/nonce guard, add the smallest per-participant/per-connection bound necessary and tests.

Do **not** implement speculative anti-cheat or a large global rate framework. If current `connectionNonce` / replacement semantics already prevent the claimed amplification, document the proof in the PR instead of changing working code.

### 7. Preserve lifecycle semantics

Do **not** solve cost control by changing the product into a short-lived account/session system.

Keep:

- anonymous Rooms;
- long-lived occupied Rooms when participants are legitimately present;
- current reconnect semantics;
- WebSocket Hibernation for idle browser/Agent connections;
- empty-room expiry/cleanup.

Do **not** add a blanket 24h Room lifetime in this PR. The problem is billable awake time and unbounded state/admission, not that an occupied Room may exist for a long time.

## Explicit non-goals for this PR

Do not add:

- SQLite DO migration;
- accounts/login/payment;
- generic Auth DO;
- Room App launch capability work from Lab #110;
- a global Turnstile page gate;
- Turnstile for native Agents;
- Cloudflare Access;
- custom distributed quota infrastructure;
- a repo-wide RoomSession refactor;
- generic WAF management code;
- unrelated cleanup.

## Testing / evidence required

Keep the current Turnstile matrix and add focused regression coverage for the combined hardening:

1. production fresh Human Turnstile + fail-closed behavior remains correct;
2. reconnect still avoids Turnstile;
3. normal resident Agent event waiting uses the hibernatable path and does not rely on a continuously chained long-poll timer;
4. any retained HTTP long-poll fallback is bounded/disabled by default and cannot keep a Room awake indefinitely;
5. Room participant maximum is enforced, while reconnect at capacity works;
6. primary serialized Room value cannot exceed its defined byte budget through legal mutations without a controlled rejection;
7. migrated expensive admission routes use the Workers Rate Limiting binding rather than KV read/write counters;
8. invalid/random pre-auth probes that receive a new cheap guard are rejected before a DO call — prove this with mocks/counters where practical;
9. existing Human Room, Agent Runtime/MCP, Room App host, voice/media and collaboration tests remain green;
10. no global Turnstile gate returns.

Run the full existing quality floor plus the relevant Room / Agent / Room-App E2E suites. If the Runtime/Harness must change to use the hibernatable event stream, update it in this same PR and run its Go tests as well.

## PR organization

**Do not open PR 2/3/4 for the items above. Extend PR #407.**

It is acceptable for this PR to touch multiple files because these changes enforce one coherent security/cost invariant. Keep commits logically separated inside the same PR if that helps review.

Only stop and propose a separate future issue for something that is genuinely a migration/architecture project (for example SQLite-backed DO storage), not for the hardening items above.

Before final review, update the PR title/body so it no longer says `PR 1 only`; summarize the complete cost-abuse hardening and clearly distinguish what is fixed from what remains intentionally out of scope.

Still stop before merge and report the exact new head, changed files, test/E2E evidence, and any production operational prerequisite such as `TURNSTILE_SECRET_KEY`.

madawei2699:
Extended in place — no follow-up PRs. New head **`837688e`** (`837688e` docs, `e1271fe` rate limiting, `8c44516` RoomSession bounds, `1017409` Turnstile).

Everything from the seven required additions, with the evidence you asked for:

1. **Turnstile work kept**, plus the one contract gap that actually existed: the widget now renders `action: "sfu-session"` and Siteverify must echo it. Absent action is tolerated (pre-#406 client builds), mismatched fails closed. Hostname validation deliberately not added — the sitekey is already domain-locked and `www`/apex/localhost are all legitimate here.
2. **Duration pinning removed at the cause**, not metered: `wait_for_events` returns the current snapshot immediately by default (still a lease heartbeat) with `longPoll` + `retryAfterMs`; the held path is an explicit `MCP_LONGPOLL_ENABLED` opt-in clamped to 5 s with a 15 s per-Room idle gap. Normal resident delivery was already the hibernatable `/api/room/agent-events` WebSocket, so **no Go change was needed** (`git status agent/` clean; the Go daemon tests already assert the official Runtime never calls public `wait_for_events`). Regression probe: default `agent-wait(25s)` leaves `vi.getTimerCount() === 0` and no parked waiter; the opt-in's clamp and gap are asserted.
3. **Participant cap 32**, shared by Human and Agent admission → `409 room_full`; duplicate still `participant_exists`; reconnect at capacity tested (and it never consults the cap). Largest existing single-Room fixture was 9, so nothing legitimate is affected.
4. **80 KiB UTF-8 serialized preflight** before `storage.put` (backend limit 128 KiB). Bounded history evicts oldest-first so chat stays writable; anything eviction cannot free → shaped `507 room_state_budget_exceeded` / WS error frame with the stored state untouched (byte-compared against a post-load snapshot); a legacy oversized record stays readable and self-heals on the next mutation instead of wedging every load.
5. **KV counters replaced by bindings** on `/api/sfu/session`, `/api/sfu/agent-session`, MCP `create_room`/`join_room` (same budgets, keyed IP + operation class; tests assert KV get/put are not called), plus new pre-DO guards for `room_info` and `/api/sfu/ws` → `429` with **zero DO invocations** for abusive probes, normal path unaffected. Stated as coarse/per-location/not-a-quota in code, docs and PR body; no authenticated-action IP limiting.
6. **Amplification bounded where the code proved a gap**: Human sockets were uncapped and older sockets stayed fully able to mutate state → now 4/participant (older replaced) + 64/Room; messages now 300/participant/10 s. The Agent event stream already self-replaces by tag, so it was documented, not changed.
7. **Lifecycle untouched**: no 24 h lifetime, no account model, hibernation + empty-room expiry + reconnect semantics unchanged.

Evidence: `yarn test` 96 files / 972 tests, `type-check`, `lint --max-warnings 0`, `prettier --check`, `docs:check`, `cf-build`, `git diff --check` all green; `wrangler deploy --dry-run` shows all three `Rate Limit` bindings; `e2e:room` 1 passed (real Worker + DO, hard-asserts `/sessions/new` was served) and `e2e:room-app-host` 6 passed. `go test ./...` is not a gate (zero Go changes); the packages that build in this sandbox pass, the rest fail at *setup* on a module cache outside the sandbox workspace.

Still stopped before merge. The one operational prerequisite is unchanged and **unconfirmed**: `TURNSTILE_SECRET_KEY` must exist as a Worker secret on `free4chat-realtime` (`cd app && npx wrangler secret list`) before the first deploy of this branch, or fresh Human joins fail closed with `503 turnstile_not_configured`. Also note the intentional default change: with `MCP_LONGPOLL_ENABLED` unset, `wait_for_events` returns immediately — the official Runtime is unaffected, third-party MCP clients should honor `retryAfterMs`.

madawei2699:
Closed in place — same PR, new head **`ee72ef2`** (`837688e..ee72ef2`). No Go change, no signed handles, no Auth DO, no accounts.

**1. Every handle-based ingress is now guarded pre-DO.** You were right that only two routes were protected. `participantHandle` is base64url JSON, not a signature, so `decodeHandle` proves nothing and the token is only checked inside RoomSession — any handle tool could resolve an attacker-supplied `roomId` into a DO invocation first.

* One shared helper (`guardedRoomControl`) now wraps **all 13** `decodeHandle() → roomControl()` call sites: `read_room_context`, `wait_for_events`, `send_text`, `update_capabilities`, `update_runtime_host`, `send_collab_request/response/result`, `send_attachment`, `publish_surface`, `clear_surface`, `read_surface`, `read_attachment`, `publish_live_view`, `leave_room`. A new tool can't forget it because the guard is the only path that resolves a Room for a handle.
* New `MCP_HANDLE_RATE_LIMITER` binding (240/60s, keyed by caller + operation class — never by Room identity, which the caller controls). `room_info`/`/api/sfu/ws` keep `ROOM_PROBE_RATE_LIMITER`; `create_room`/`join_room` keep `MCP_JOIN_RATE_LIMITER`.
* Our own rejection uses a distinct name (`mcp_ingress_rate_limited` → tool error `rate_limited`) so a DO 429 such as `activity_capacity` keeps its own meaning.
* Tests: a forged handle on `wait_for_events`, `send_text` and `read_surface` returns `{error:"rate_limited"}` with **zero DO invocations** (recording DO namespace); a legitimate handle still reaches its Room. The "rejected before any DO call" claim now holds for every pre-auth Room resolution.

**2. `wait_for_events` cadence is enforced by the server.** Two layers, neither client-dependent:

* **Pre-DO**: `MCP_WAIT_RATE_LIMITER` (1 per 10s per caller + participant) refuses with a shaped `{ "error": "wait_rate_limited", "retryAfterMs": 10000 }` before any Durable Object is woken — a tight loop no longer even reaches the DO.
* **In RoomSession**: the same floor (5s) is enforced authoritatively per participant, checked *before* the lease refresh and `saveRoom`, so the refusal cannot be used to keep a Room busy or to amplify Room-value writes. Tests assert the refused call does not change `lastSeenAt` or stored state, and that the next call past the floor is accepted.
* The resident Runtime is untouched: it uses the hibernatable `/api/room/agent-events` stream, and the existing Go daemon tests assert the official Runtime never calls public `wait_for_events`.

Four existing tests that assert two sequential waits (`replay` dedup, Human-Task scoping, the two attachment-half wake tests) now advance a mocked clock instead of hammering the same participant; their own invariants are unchanged.

**Evidence on `ee72ef2`**: `yarn test` 96 files / **977 tests**; type-check, `lint --max-warnings 0`, `prettier --check .`, `docs:check`, `git diff --check` green; `cf-build` + `wrangler deploy --dry-run` show **five** `Rate Limit` bindings and all new error strings in the bundle; `e2e:room` 1 passed and `e2e:room-app-host` 6 passed. CI re-running now.

Still stopped before merge. Unchanged operational prerequisite: confirm `TURNSTILE_SECRET_KEY` exists as a Worker secret on `free4chat-realtime` (`cd app && npx wrangler secret list`) — I cannot see Cloudflare secrets from here. New client-visible behavior: a third-party MCP client polling `wait_for_events` harder than once per 10s now gets `wait_rate_limited` and must honor `retryAfterMs` (the official Runtime and the browser are unaffected).

PR #413, opened 2026-09-19
feat(room-app): scale the in-Room App launcher (Lab #98)

madawei2699:
## Lab issue

`i365dev/free4chat-extension-lab#98` — **architecture(room-app): scale the in-Room App launcher without making Apps dominant** (ACTIVE: 18 promoted runtimes in the production catalog).

Core PR only. **Do not merge automatically.** The Lab catalog/portfolio ownership is untouched.

## What the current implementation was

- `app/src/components/RoomContent.tsx` rendered the Stage entry as a `role="tablist"` (`data-testid="stage-switcher"`) that mapped **every** curated App to one `stage-app-<id>` chip inside a horizontal scroller (`overflow-x-auto`), followed by the existing `Screen` / `Live View` chips.
- The catalog came from the Lab-owned control plane via `loadProductionRoomAppCatalog()` (`app/src/common/roomApp.ts` → `https://room-apps.free4.chat/_catalog.json`, bounded v1 schema, 60s refresh, fail-closed).
- Live production catalog today: **18 active Apps** (`_catalog.json` fetched during this work). At that size the strip hides most Apps off-screen on desktop and cannot be navigated at 390px.

No missing catalog field was found: `id`, `label`, `path`, `status` are sufficient for this UI, so the Lab seam is unchanged (see *Contract*).

## Chosen UX (smallest progressive disclosure)

```text
Stage strip:  [current App] [recent App] [recent App] [Apps…]     (desktop: ≤3 App chips)
              [current App] [Apps…]                                (~390px: 1 App shortcut)
              Room conversation stays untouched beside it

[Apps…] → lightweight popover (desktop) / viewport-fitted surface (phone):
           Search apps…
           Recent in this Room   (this browser tab, this Room, most recent first)
           All apps              (canonical catalog labels, catalog order)
           Explore all apps →    (explicit link, existing /apps discovery)
```

- Nothing is rendered inline until an App is actually opened; no default/fake recent set.
- The current App is always the first inline chip and is marked (`●` + `aria-pressed`), so "current" stays truthful even if a chip is truncated.
- Selecting an App closes the launcher and reuses the existing host lifecycle (`launchRoomApp` + resident host + `fullscreen` semantics unchanged).
- `Room` remains reachable: the conversation pane is never replaced, closing an App is the existing one-click path on the host chrome, and the strip stays a single local scroller.

## State location for recency

`app/src/common/roomAppRecents.ts` — pure, bounded helpers plus:

- **Storage:** `sessionStorage`, key `free4chat:room-app-recents:v1:<roomName>` — per browser tab, cleared when the tab closes, namespaced per Room name. **Never `localStorage`**, no server state, no accounts.
- **Bounds:** two explicit, independent product limits — `ROOM_APP_INLINE_SHORTCUTS_MOBILE = 1` and `ROOM_APP_INLINE_SHORTCUTS_DESKTOP = 3` — passed as an App-shortcut budget that `Apps…` never consumes; remembered list ≤ 6; de-duplicated (reopening only reorders); pruned against the current catalog so a retired App silently drops out.
- **Hydration order:** read this Room's recents → mark the Room hydrated → prune → persist. Nothing writes, prunes, or reorders before the Room is hydrated, and the prune basis requires a catalog the Room can actually offer, so neither an in-flight catalog nor a transient `roomAppsEnabled=false` (media reconnect) can be misread as "every remembered App was retired".
- **No ranking, favorites, categories, accounts, or cloud preferences** — recency + canonical catalog order only.

## Files changed

| File | Change |
| --- | --- |
| `app/src/components/RoomAppLauncher.tsx` (new) | Search / recent / all-apps launcher, keyboard support, desktop popover + phone surface |
| `app/src/common/roomAppRecents.ts` (new) | Bounded Room-scoped recency + inline-set helpers (sessionStorage only) |
| `app/src/components/RoomContent.tsx` | Stage strip now renders the bounded inline set + `Apps…`; recency wiring |
| `app/src/components/RoomAppLauncher.test.tsx` (new) | Launcher regressions |
| `app/src/common/roomAppRecents.test.ts` (new) | Recency regressions |
| `app/src/components/RoomContent.test.tsx` | Adapted strip tests + 9 issue-scoped integration tests |
| `app/e2e/room-app-host/room-app-host.spec.ts` | Compatibility gate goes through the real launcher; catalog reachability + search + keyboard |
| `app/e2e/fixtures/room-app-fixture.ts` | Fixture catalog is now 5 active Apps (was 1) so "the strip is bounded, the launcher is not" is a real assertion |
| `app/e2e/room-app-host/launcher-smoke.spec.ts` + `launcher-smoke.playwright.config.ts` (new) | Non-CI visual smoke: screenshots + printed geometry at 1280×800 and ~390px |

## Browser observations (desktop + ~390px)

Chromium smoke run, printed as data (screenshots written to `test-results/launcher-smoke/`):

```text
desktop-launcher  viewport 1280x800  pageOverflow 0
  strip: local scroll, 1 row          launcher 384x287 @ (8,89), inside viewport
  all 5 catalog rows listed, uniform width 366
desktop-app-open  viewport 1280x800  pageOverflow 0
  inline chips: ["stage-app-core-fixture-app"] (1)   strip rows: 1
phone-launcher    viewport  390x844  pageOverflow 0
  strip: local scroll, 1 row          launcher 374x279 @ (8,158) — directly under `Apps…`, inside viewport
```

All six Playwright viewport projects pass (chromium-desktop, webkit-desktop, webkit-phone 390×844, iPad portrait/landscape/1024). No page-level horizontal overflow in any state.

## Tests

```text
yarn type-check   clean
yarn lint         clean (0 errors, 0 warnings)
yarn test         99 files / 1071 tests passed
e2e room-app-host 6/6 projects passed
smoke             desktop reload keeps this tab's recents; phone with 3 Apps open shows 1 shortcut
```

New regressions cover: all promoted Apps discoverable · bounded inline recent set · opening an App updates recency · duplicates do not accumulate · current App visible · search finds Apps · search empty state · Room remains reachable/no dead chip · no persistent account/favorites behavior (asserts `localStorage` stays empty) · keyboard open/search/select/close · catalog-maximum (64 char) labels do not break layout · **pre-seeded recents survive mount, unmount/remount in the same tab, per-Room independence, and a transient `roomAppsEnabled=false`** · **exactly one mobile App shortcut plus `Apps…` after opening 3+ Apps at 390px**.

## Contract

No Lab/Core contract blocker. Recency is deliberately **Core-local and generic**; the launcher consumes only the existing bounded catalog fields and adds no Lab-side requirement. If the Lab later wants a canonical "recent" signal or category metadata, that would be a separate, explicit catalog-version change — not needed for this fix. The success criterion holds: 30+ Apps stay reachable without the strip growing.

## Risks / notes

- The inline strip no longer shows an unopened App, so a deep link (`/room?...&app=<id>`) opens inline immediately (unchanged) while a cold Room shows only `Apps…`. This is the intended progressive disclosure; the e2e compatibility gate was updated accordingly rather than being bypassed.
- Recency is per browser tab. A second tab in the same Room starts with an empty recent set by design (no cross-device, no server state); a reload in the same tab keeps it, which is now proven in the browser smoke, not just by the storage helper.
- At ~390px the strip is `[current App] [Apps…]` (or `[most recent App] [Apps…]` with nothing open): exactly one App shortcut, never two.
- `launcher-smoke.*` is intentionally **not** wired into CI — it exists for human visual review; the semantic gate remains `room-app-host.spec.ts`.


madawei2699:
Review on current head `60d8e681396f065ed058415e9d827e53528fa9fd`.

The overall #98 direction is right and the Core/Lab boundary is clean, but I found **two concrete regressions to fix before merge**. Both are small and local; please do not widen scope.

### 1. sessionStorage recency is erased before it is rehydrated

In `RoomContent.tsx`, the effects currently run in this order:

1. persist `prunedRecentRoomAppIds`;
2. then read `readRecentRoomAppIds(roomName)`.

On a fresh mount, React state starts as `[]`, so the first effect calls:

`writeRecentRoomAppIds(roomName, [])`

which removes the existing sessionStorage entry. The following rehydrate effect then reads an already-cleared key.

So the documented/reported behavior:

`reload same Room in same tab -> recent Apps survive`

does not actually hold, and the current tests only test the storage helper round-trip, not the component mount/reload path.

Please fix the lifecycle so **read happens before any write for that Room**, e.g. with an explicit hydrated/loaded-for-room guard. Avoid SSR/hydration mismatch; do not simply read browser storage in render if that changes first paint.

Required regression:

- pre-seed `free4chat:room-app-recents:v1:test-room` with two valid App ids;
- mount `RoomContent`;
- prove the stored ids are not deleted during initial effects;
- prove the inline/launcher recency reflects them after hydration;
- reopen/remount the same Room in the same tab and prove they survive.

Also cover changing to a different Room name if the component supports that path: Room B must not inherit Room A.

### 2. mobile bound does not match the shipped UX claim

The report says:

`~390px strip: [current App] [Apps…] (≤2 chips)`

but `ROOM_APP_INLINE_RECENT_MAX = 2` is passed as the **App-id count**, and `Apps…` is added separately. After opening multiple Apps on a phone, this can render:

`[current App] [recent App] [Apps…]`

= 3 controls, not 2.

For the intended mobile progressive-disclosure surface, make the phone App-id bound **1**:

`[current App] [Apps…]`

or, with no active App:

`[most recent App] [Apps…]`

Desktop can remain:

`[current] [recent] [recent] [Apps…]`

Do not couple the desktop bound to `mobile + 1` if that forces the wrong number; make the product limits explicit.

Add a 390px/component regression after opening at least 3 Apps that proves only one App shortcut plus `Apps…` is inline.

Everything else can stay as-is. Re-run the same unit / typecheck / lint / Room App host matrix / browser smoke, then report the new head SHA. If these two are green, this should be merge-ready.

madawei2699:
Both blockers are fixed on head `fa4c1713730966ac391278ecd67d3668467373cc` (rebased onto the current `cf-sfu`; the two new base commits were unrelated — task attachments and ACP session primitives — with no overlap).

### 1. Recents are hydrated before anything can persist or prune them

`RoomContent.tsx` now runs the lifecycle in one explicit order, with the read kept in an effect (never in render, so SSR/first paint still never touch browser storage):

```text
read this Room's recents  →  mark that Room hydrated  →  prune  →  persist
```

Three guards make it hold:

1. **Nothing writes, prunes, or reorders until the Room has been hydrated** (`recentsHydratedRoom === roomName`). The open-App recency effect waits too, so a deep-linked App cannot reorder a list that has not loaded.
2. **The prune basis requires a catalog this Room can actually offer** (`roomAppsEnabled && roomAppCatalogLoaded`). Both an in-flight catalog and a transient `roomAppsEnabled=false` (ordinary SFU/media reconnect) present an empty available-App set; that must not read as "every remembered App was retired".
3. **Prune and persist are one effect that re-derives the value it writes.** Splitting them left a commit where the persist effect still saw the previous render's derived list.

Guard 3 and the availability half of guard 2 were found only by driving a real browser: after adding guard 1, the Playwright smoke still showed the key being deleted on reload, and instrumenting `sessionStorage.removeItem` plus the persist decision identified the two stale/transient writes precisely. That is why the smoke now asserts reload persistence for real instead of relying on the storage helper.

Required regressions (all mutation-checked by reintroducing the original lifecycle — 10 tests fail against it):

- pre-seeded `["test-app-4","test-app-2"]` is **not** erased during initial effects; the inline strip and the launcher's "Recent in this Room" both reflect it afterwards;
- unmount/remount of the same Room in the same tab keeps them;
- Room B does not inherit Room A, keeps its own storage, and leaves Room A's entry untouched;
- opening a new App after hydration persists the updated order, and reopening an already-recent App only reorders (no duplicates);
- a transient `roomAppsEnabled=false` does not erase storage and the order returns unchanged;
- an unbound Room (`roomName=""`) reads nothing and erases nothing.

### 2. Mobile is one App shortcut plus `Apps…`

The bounds are now two explicit, independent product limits instead of one derivation:

```text
ROOM_APP_INLINE_SHORTCUTS_MOBILE  = 1     // [current App] [Apps…]
ROOM_APP_INLINE_SHORTCUTS_DESKTOP = 3     // [current] [recent] [recent] [Apps…]
```

`inlineRoomAppIds` takes an App-shortcut budget that `Apps…` never consumes, and `max` lost its default so no caller can silently pick a bound. At 390px with an App open: `[current App] [Apps…]`; with nothing open: `[most recent App] [Apps…]` — never `[current] [recent] [Apps…]`.

Regressions (mutation-checked against `MOBILE = 2`, which fails them): opening 3 Apps at mobile width leaves exactly one inline shortcut, `Apps…` present, and the full catalog still in the launcher; and the no-current-App case shows the single most recent App. The browser smoke additionally opens three Apps on a 390px page and reports `inlineChips: ["stage-app-fixture-third-app"]`, one strip row, zero page overflow.

### Validation on `fa4c171`

```text
type-check        clean
lint              clean (0 errors, 0 warnings)
prettier          clean
git diff --check  clean
unit              99 files / 1071 tests passed
Room App host compatibility matrix   6/6 viewport projects passed
browser smoke     passed — desktop reload keeps recents; phone w/ 3 Apps = 1 shortcut
```

GitHub CI on `fa4c171`: App Tests ✅, Lint & Type Check ✅, Format & lint autofix ✅, Room App host matrix ✅, gofmt/vet/test/build ✅, Distribution/cleanup ✅, Build & Deploy skipped (not `cf-sfu`). `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`. PR description updated so the published bound claim now matches the implementation.

Not merged, per the request. No remaining blocker on my side; the only follow-up I am aware of is the third-party code-review bot, which was rate-limited on this PR rather than reporting findings.

Issue #409, opened 2026-09-18
enhance(task): lightweight remote Task control for resident Agents


madawei2699:
## Status

**PRODUCT / RUNTIME ENHANCEMENT — DOGFOOD-DRIVEN. DO NOT TURN THIS INTO A CHATGPT/IDE CLONE.**

Canonical boundary: #51

## Why this exists

Free4Chat already has most of the primitives needed to keep a local Harness Agent in a temporary Room: resident Go Runtime, ACP Harness, retained Task scopes, Task text/attachments, coarse activity, permission presentation when supported, artifacts, and Live View.

But the current Task experience still feels experimental for real long-running coding/research work. The practical workflow we want is:

```text
large task brief
→ local Agent starts a long task
→ Human leaves the computer
→ phone / another browser can tell whether the task is alive
→ Human can send a small follow-up
→ Human can interrupt the current turn
→ Human can answer a permission request when the Harness supports it
→ Human can continue interacting with the same Task while its Harness session survives
→ final result / artifacts remain easy to inspect
```

This is not an attempt to reproduce the full Codex / Claude / ChatGPT UI. The goal is a **lightweight, cross-Harness remote Task control surface**.

## Current evidence / problem

- Task status is mainly derived from retained Room messages/collaboration lifecycle.
- A Human cannot clearly distinguish healthy long-running work, queued follow-up, Runtime reconnect, Harness/session loss, and interruption.
- Agent activity is intentionally coarse and does not yet provide enough confidence during long work.
- Follow-up instructions can arrive while work is running, but the UI does not clearly say whether they are handled now or queued.
- The Runtime currently serializes addressed Harness execution through one global `turnRunning` drain. Multiple retained Task scopes provide context separation, not guaranteed concurrent execution.
- The ACP adapter already has `session/cancel` / `CancelTurn()`, but Room Task UX does not expose interruption as a first-class control.
- Harness permission support is uneven. Free4Chat should present native approvals when available, but must not pretend every Harness implements them.
- Real Harnesses may exit or lose their ACP session. This must fail visibly rather than leaving a Task looking indefinitely `Working`.

## Product model

> **Task = one remotely observable and controllable unit of Agent work inside a temporary Room.**

Conversation is one Task channel, not the whole Task product.

```text
Task
├─ initial brief / context
├─ retained Harness scope
├─ execution state
├─ bounded activity / health
├─ follow-up instructions
├─ optional permission requests
├─ artifacts / result
├─ interrupt
[...]

PR #417, opened 2026-09-19
feat(task): hand off an existing Pi session into a Task (#409)

madawei2699:
### PR: #417 — `feat(task): hand off an existing Pi session into a Task (#409)`

**HEAD:** `959ea54d3601f416f96e9bb5e049aaa3569a7c73` (branch `codex/409-pi-session-handoff`, base `cf-sfu @ f87cb3678f2ebcbcfb5d27f3c59ddb83ca109919`)

**handoff model:** cold resume / cooperative handoff only. The operator works several turns in the native Pi CLI, stops (the Pi turn must already be idle), then from a normal local shell:

```
free4chat-agent handoff --list [--cwd <path>] [--cursor <token>] [--instance <id>]
free4chat-agent handoff --adopt <session-id> [--human <participant-id>] [--cwd <path>] [--instance <id>]
free4chat-agent handoff --status | --clear [--instance <id>]
```

`--adopt` arms exactly ONE pending adoption in the resident Runtime's memory and touches nothing else. The Human then creates the next normal Task in the Room; that canonical `task:<requestId>` scope is bound to the selected native session, and the first Task instruction continues the native Pi conversation. Follow-up messages, queueing, interrupt, Interrupt & send, permission handling, and the #416 execution projection are all the already-landed behavior and are unchanged. Explicitly NOT in this slice: hot takeover, cross-client cancel, session stealing, steering, mid-turn injection, multi-Task concurrency, non-Pi handoff, a browser session picker, any new transport/server/DB, session persistence, and any lease/broker abstraction.

**Pi admission:** admission is a named-product decision, not a capability check. Only a resident whose launcher is named `pi` **and** whose adapter actually implements the session primitives (`ListSessions`/`LoadSession`, landed in #412) is eligible; everything else fails with `session handoff is supported for the Pi Harness only`. The ACP-advertised `sessionCapabilities.list` / `loadSession` flags are deliberately NOT the gate — the earlier spike showed OpenCode and Hermes advertise them and still cannot continue a native session. The Runtime seam is an unexported structural interface, so `types.HarnessAdapter` and every other adapter/fake are untouched.

**CLI:** `handoff` requires exactly one of `--list` / `--adopt <id>` / `--clear` / `--status` (anything else is the usage error, exit 2). It routes through the existing daemon IPC boundary (`handoff-list` / `handoff-adopt` / `handoff-clear` / `handoff-state`) with the existing `--instance` resolution (sole resident by default, explicit choice when ambiguous) and prints only the daemon's bounded JSON. Nothing is created in the Room, and the session id is never echoed back into terminal output.

**daemon:** the four ops are thin passthroughs to one resolved resident: `handoff-adopt` returns `{state:"armed", armed, humanParticipantId}` (never the session id), `handoff-state` returns `{armed, humanParticipantId}`, `handoff-clear` returns `{state:"cleared"}`, `handoff-list` returns the adapter's bounded descriptors plus cursor. A failing discovery is surfaced rather than reported as "no sessions", and a non-Pi resident is rejected by name.

**Runtime adoption:** `admitHarnessSession(scope, target)` is now the single serialized admission boundary in the turn pipeline. For a Task scope it: (1) fails closed if the scope already adopted a session that is gone; (2) otherwise keeps the adapter's retained conversation; (3) otherwise binds an armed adoption when — and only when — that scope's pending trigger carries a canonical Human-owned collab request for this Agent; (4) otherwise takes the unchanged fresh-scope path. Adoption binding consumes the pending adoption BEFORE the load, so the local session identity cannot outlive the attempt, and it is consumed exactly once (two Tasks admitted before the drain produce exactly one `session/load`).

**Task binding:** binding requires canonical Human ownership, exactly like every existing Task control. With `--human <participant-id>` the adoption binds only that Human's Task; without it, only when the Room has exactly one Human and that Human owns the Task. In any other case (another Human's Task, or an ambiguous multi-Human Room) the Task keeps the normal fresh-scope path and the adoption stays armed — `handoff --clear` is the recovery path. The Room scope is never adopted, and the ordinary startup `EnsureSession()` for the Room conversation is unchanged.

**session-loss behavior:** a Task that adopted a native session never falls back to a fresh conversation. If `session/load` fails, the scope is marked adopted-and-unavailable, the adapter error is reduced to its bounded failure class in the log (never logged verbatim, since it can quote local ACP identity), and the turn fails with the fixed secret-free error `the adopted session for this Task is no longer available`. If the Harness dies later, every adopted scope is additionally marked lost. In both cases later instructions for that Task fail closed with the existing `availability: "session_lost"` projection, and the failure is now classified as **permanent**, so it cannot consume the bounded retry budget or keep re-executing the head of the serial turn queue while another scope's fresh work waits.

**normal Task compatibility:** with no adoption armed, a Task is byte-for-byte the previous behavior (one scoped `session/new`, then the turn), and a follow-up reuses that session. Regression fences cover the ordinary path, Room-scope isolation, and an unrelated Task continuing normally after an adopted Task has failed.

**session/new-before-load proof:** two independent layers. (a) Runtime: the fake adapter records a scoped `new:` only when a session is *actually created* — the adopted scope shows exactly one `load:` before its first `run:`, and zero `new:` before the load, after the load, and after session loss. (b) Adapter: against the real `ACPAdapter` with the fake ACP child, the trace shows exactly one `session/new` (the Room's), one `session/load`, and the scope's later `EnsureSessionFor` adds no `session/new` at all, while the scoped turn is addressed to the loaded session id.

**privacy boundary:** the ACP session id exists only in the operator's own CLI argument, the daemon IPC message, and the pending runtime state, and is dropped from runtime state as soon as the adapter has loaded it (after that the adapter owns the identity and the runtime keeps only "this scope is adopted"). It never enters Room state, Room messages, prompts, workspace files, analytics, or the lifecycle log; the log lines carry only `scopeKind` and the bounded failure class. Tests assert the id is absent from sent Room text, collab responses, execution projections, lifecycle logs, and `Runtime.Status()`, and from the `handoff` responses. One pre-existing, explicitly local surface is unchanged and worth naming: `free4chat-agent status` already reports retained conversations through the documented daemon/CLI-only `harnessSessions` diagnostics, so a loaded session appears there exactly like the Room's and every task-scoped session — that surface is local IPC, never Room/browser/telemetry.

**tests:** `go test ./...` green (all 13 packages). New coverage: Runtime adoption invariants (bind-next-Task with the exact session id/cwd, wrong-Human non-consumption then correct-Human consumption, unpinned single-Human rule, one-pending + clear + invalid-id rejection, exactly-one-adopter with two Tasks admitted before the drain, load failure, post-loss fail-closed with an unrelated Task still working, ordinary-Task regression, Room-scope isolation, Pi-only/primitive-less rejection, bounded local discovery); the adapter ordering trace; daemon IPC (no session id echoed, bounded descriptors, failing discovery surfaced, Pi-only/unknown-instance rejection, sole-resident default); CLI composition (exact request per mode, usage exit 2 for no/multiple/empty selectors).

**race:** `go test -race ./internal/runtime ./internal/harness ./internal/free4chat` green, and `go test -race ./internal/daemon -skip TestRuntimeExecutableCopySurvivesInstalledPathReplacement` green. That skipped test is a **pre-existing** race unrelated to this slice: it reads `d.runtimeExecutableCopy` while `Daemon.Run()` writes it in `prepareRuntimeExecutable` (`daemon_test.go:1087` vs `runtime_executable.go:75`). I reproduced the identical failure on the pristine baseline `f87cb36` in an isolated worktree, and left it alone rather than widening this PR. It also means a plain `-race ./internal/daemon` run is red on this branch and on `cf-sfu` alike.

**real Pi probe:** partially reproduced, and honestly bounded. With a locally installed Pi (`pi 0.84.4`), the real launcher (`npx -y pi-acp@0.0.33`) starts and completes a real ACP `initialize`, advertising `agentCapabilities.loadSession: true` **and** `sessionCapabilities.list: {}` — a direct reminder of why V1 does not trust that advertisement. A real native-session continuation could not be demonstrated in this sandboxed session: the bridge cannot create/load sessions because it must write its own `~/.pi/pi-acp/session-map.json` (EPERM under the workspace-write sandbox), and with a disposable HOME containing a copy of the native session store it reaches the real Pi CLI, which then refuses `session/new` with `Authentication required` because I deliberately did not copy the operator's Pi credentials into temporary storage. No Harness turn was ever run, no native conversation was created or modified, the disposable copy was deleted, and the probe file was not committed. The invariants therefore rest on the deterministic adapter/runtime/daemon/CLI tests above; the operator can rerun the same probe in a normal unsandboxed shell to see the real conversation continue.

**CI:** see the checks on this PR (App Tests, Lint & Type Check, Format & lint autofix, Room App host matrix, `gofmt / vet / test / build`, Distribution / cleanup contract, build matrix). `app/**` is untouched, so the Web jobs exercise existing behavior only.

**explicitly untouched:** `app/**` (no browser protocol change, no session id in the Room, no session picker), the Room/Durable Object protocol, `types.HarnessAdapter`, the #414 interrupt path, the #416 execution projection and UI, non-Pi Harnesses (Codex/Claude/OpenCode/Hermes/custom), session persistence, and every release artifact.

**release: NOT released** — no Agent version bump, no release, no published binaries, no production activation, and no full manual regression requested. `Refs #409` only; issue #409 stays closed and is not reopened.


madawei2699:
Follow-up commit `17360f9` on this same branch fixes three correctness blockers inside the accepted V1 architecture. No product scope, Room/browser protocol, or PR boundary changed.

**1. The adoption is now fenced to Tasks created AFTER arming.** `pendingSessionAdoption` carried an `armedAt` timestamp that was never a causal authority, so a Task accepted *before* `--adopt` could sit in the serial queue, be reached only after the running turn finished, and consume an adoption meant for the next Task. It now carries `armedAfterSequence`, captured under the existing Runtime lock as this Runtime's canonical receipt/admission boundary: the monotonic Room receipt cursor, the bounded event buffer (covers an envelope that is mid-ingestion before the cursor advances), and every already-admitted canonical sequence per logical scope — no wall clock, no new counter. A Task at or below the fence runs its completely normal fresh-session path and **leaves the adoption armed**; the first eligible Task above the fence consumes it. `--human` / single-Human ownership still applies *on top of* the fence.

**2. Adopted ownership is committed BEFORE the external `session/load`.** Previously the scope was marked adopted only after the load returned, so a Harness death inside the load window was invisible to the failure boundary: nothing recorded the loss, and the next admission was free to create a fresh scoped conversation — silent replacement, the one outcome #409 must never allow. Consuming the pending identity, marking the scope adopted, and clearing any impossible stale lost bit now happen in one critical section *before* `LoadSession`. Load failure keeps the existing fail-closed semantics; load success now checks whether the scope became lost while the load was in flight and, if so, returns `the adopted session for this Task is no longer available` without starting the first Task turn and without erasing the loss.

**3. "New conversation" and "needs the Free4Chat bootstrap" are now separate facts.** The first adopted turn was reported as `New`, so the prompt literally said *"This is a new local Harness session."* while continuing an existing native Pi conversation. `HarnessSessionContext` now carries `Bootstrap` beside `New`, and the Runtime derives both from the **existing** generation markers (`newSession` stays true until a Harness delivery for that generation is acknowledged), so no second bootstrap state machine exists:

| turn | `New` | `Bootstrap` |
| --- | --- | --- |
| adopted first turn | `false` | `true` |
| adopted follow-up (after acknowledgement) | `false` | `false` |
| adopted first turn retried before acknowledgement | `false` | `true` |
| normal fresh Task | `true` | `false` |

The prompt then teaches the same authority/reply/affordance contract, adds the truthful line *"This is the first Free4Chat-controlled turn in an existing local Harness session. Current Room sequence: N."*, keeps the bounded pull-only Room-context notice, and never claims the conversation is new.

**Identity cleanup:** an opaque ACP session id is no longer trimmed before arming (accepted unchanged or rejected explicitly), the CLI no longer repairs it either, and `cwd` reaches `session/load` byte-for-byte. Final identity validation stays with the #412 adapter.

**Deterministic tests** (no sleeps; each verified to fail when its own fix is reverted):
- the exact pre-arm scenario — Task B `seq=10` queued before arming → fresh `session/new`, no `session/load`, adoption still armed; then Task C `seq=11` after arming → `session/load native-pi-1`, no scoped `session/new`, adoption consumed — plus a white-box assertion that the captured fence already covers an admitted-but-undrained Task;
- a Harness death fired from inside `LoadSession` before it returns success → scope is adopted **and** lost, first Harness turn does not run, no scoped `session/new`, later instructions stay `session_lost` with no replacement `RunTurn`;
- adopted-first / acknowledged-follow-up / ordinary-Task / retry `Session` semantics captured off the real turn pipeline, and the rendered prompt asserted to carry the bootstrap contract + truthful existing-session wording and never the new-session line;
- identity preservation at the Runtime and CLI seams (padded id rejected, cwd with spaces passed through unchanged).

**Validation on `17360f9`:** `gofmt -l .` clean; `go vet ./...` clean; `go test ./... -count=1` green (13 packages); `go test -race ./internal/runtime ./internal/harness ./internal/free4chat` green; `go test -race ./internal/daemon -skip TestRuntimeExecutableCopySurvivesInstalledPathReplacement` green (that skipped pre-existing race is reproduced on the clean baseline and deliberately left out of this PR); `go build ./cmd/free4chat-agent` ok; `git diff --check` clean. No version bump, no release, no production action. The earlier real-Pi sandbox limitation is unchanged and still stands — no credentials were copied and no sandbox boundary was weakened.

Issue #421, opened 2026-09-19
enhance(task): long-lived, resumable and bounded-concurrent Agent execution


madawei2699:
## Why this exists

Dogfood around #409 / #420 exposed a Runtime execution-model limit that is now a real product problem rather than an abstract optimization.

Free4Chat increasingly treats a Task as a retained Agent work session. Real coding/Agent workflows may run for tens of minutes or several hours, while the Human may leave the Room and return later.

Today one resident Agent is effectively a **single global execution lane**:

~~~text
Resident Agent
├─ Task A / session A
├─ Task B / session B
└─ Task C / session C

ResidentRuntime
  turnRunning = one global turn

ACPAdapter
  promptActive = one global prompt

=> A blocks B and C
~~~

This creates three related product problems:

1. **Long-lived work**
   - ACPAdapter currently defaults TurnTimeoutMs to 120000 ms (2 minutes).
   - Production can override it with FREE4CHAT_ACP_TURN_TIMEOUT_MS, but a fixed short timeout is not an acceptable product model for real Agent tasks.
   - Real sessions may legitimately work for 20 minutes, 1 hour, or several hours.
   - A timeout should recover a truly stuck Harness, not define the maximum useful Task duration.

2. **Resumable supervision**
   - The Human should be able to leave the Room/browser while the local Agent continues working.
   - Later the Human may re-enter, reopen the Task, inspect its current/coarse state, continue the same retained Harness session, interrupt it if still running, or send follow-up work.
   - Continuous high-frequency live monitoring is NOT required for correctness.

3. **Head-of-line blocking across independent Tasks**
   - ResidentRuntime.turnRunning serializes all scopes for the whole resident.
   - ACPAdapter.promptActive serializes all prompts for the whole ACP connection.
   - Therefore Task B cannot make progress while unrelated Task A is running, even when A and B are different retained Harness sessions.
   - Session Picker / existing-session continuation makes this much more visible because users will naturally start multiple independent Tasks against different sessions.

Refs #51, #409, #420.

---

## Product target

A Free4Chat Agent participant should support:

~~~text
long-lived local Agent work
+ Human may disconnect/reconnect
+ retained Task/session can be resumed
+ independent Tasks can make bounded concurrent progress
~~~

Example:

~~~text
Task A / Session A
  running for 1h+
  Human leaves Room

Task B / Session B
  can still start if execution capacity exists

later:
Human rejoins Room
→ opens Task A
→ sees truthful coarse state
[...]

PR #422, opened 2026-09-19
feat(task): long-lived, resumable and bounded-concurrent execution (#421)

madawei2699:
Refs #421 — **not** `Fixes`, because the issue's success criteria include a final production dogfood that this PR deliberately does not perform.

## What this changes

A retained Task is now a real work session: it may run for hours, survive the Human leaving the Room, resume behind an unrelated Task, and be interrupted exactly.

**No release, no version bump, no tag, no bootstrap activation.** `agent/internal/doctor/doctor.go`, `app/public/agent.md`, `.github/`, and every release script are untouched.

## 1. Provider evidence first (real probes, not inference)

Every pinned bridge was driven over **raw ACP** by a standalone probe: one bridge process, two independent `session/new` sessions, session A parked in a long shell tool call, then session B submitted while A was still unsettled. Classifications:

| Provider | Concurrency | Routing | Isolation suite | Classification |
|---|---|---|---|---|
| **Pi** | 3/3 trials: B settled in 1.2–2.6s while A ran 45–60s, with A's own notifications arriving *strictly inside* B's window | correct both ways | conversation memory isolated; `session/cancel` on A settled A in 2.2s while B ran on to completion; killing one per-session `pi` worker left the bridge alive and a brand-new session working | **VERIFIED CROSS-SESSION** |
| Codex | concurrent progress observed | correct | not run | concurrency-observed → **SERIAL** |
| OpenCode | concurrent progress observed | correct | not run | concurrency-observed → **SERIAL** |
| Hermes | concurrent progress observed (71 interleaved A updates) | correct | not run | concurrency-observed → **SERIAL** |
| Claude | — | — | — | **UNVERIFIED**: local bridge credentials expired, so every prompt failed to authenticate. An errored prompt is not evidence about concurrency. |

Only Pi is enabled for cross-session execution, and Pi is also the only Harness with #420 Task Session Continuation — exactly the workflow that makes two independent retained sessions normal. The evidence and its limits live next to the launcher, in the ONE centralized `TaskExecution` policy (`types.AgentLauncher.TaskExecution`); nothing branches on a Harness identity anywhere else (`grep -rn '"pi"' agent --include=*.go | grep -v _test` → the launcher registry line only).

## 2. Execution model: multiplex, not pool

The ACP adapter already owned per-scope session routing (`sessions[scope]`, scope-resolved activity and permissions). The only genuinely global state was five prompt fields. So:

* active turn state is keyed by **native session id** — the strongest identity available, which keeps *one conversation, one turn* enforced at the only layer that knows session identity, even if two logical scopes somehow named the same session;
* `CancelTurnFor(scope)`, `TurnOwnerFor(scope)`, and per-conversation permission sets are the exact-control seams;
* a second turn for a busy conversation returns a typed **deferral**, not a failure.

Rejected alternative: a bounded pool of isolated adapters. It costs strictly more RAM (each bridge process is its own ~180 MB+), needs lane assignment plus discovery/load/capability routing per lane, and buys isolation that the probe showed Pi already provides per conversation. One process multiplexing N conversations is the minimum structural change — *and the probes are what made that safe to claim rather than assume*.

## 3. Bounded concurrency in the Runtime

`turnRunning bool` is replaced by a bounded lane set. Candidate scopes are visited in the same fair order the serial drain always used (Room scope first, then scope admission order), so a long-lived Task occupies **one lane**, never the queue:

```
capacity available:  A -> Running, B -> Running
capacity full:       C -> Queued   (truthful, never an apparent hang)
```

* a scope never executes two turns at once;
* a scope bound to a conversation another scope is executing is deferred, not dispatched only to be refused;
* a failed turn ends *its own* scope's drain pass — it is never relaunched in a tight loop, and it never blocks an independent scope in the same pass;
* the autonomous retry budget is keyed per **canonical turn**, so two concurrent failures can no longer reset each other's bounded budget (a real unbounded-retry hazard the single-slot bookkeeping would have introduced).

`MaxTurnLanes = 2` is a hard product ceiling; any missing, malformed, or larger policy value fails safe to 1 (the exact pre-#421 behavior), proven by table test.

## 4. Long tasks: lifetime separated from stuck recovery

The 2-minute default was a product cap disguised as a safety bound. A turn now has:

* **process liveness** — a dead child is detected immediately, no timer involved;
* a deliberately conservative **6-hour safety ceiling** (env-overridable as before);
* **explicit Human interrupt**, unchanged and now lane-exact;
* an **OPT-IN idle watchdog**, off by default *because the probe measured it*: a 60-second shell command produced **zero** notifications between `tool_call` and its completion, so a default idle bound would kill legitimate long tool work. When enabled it is re-armed by real provider activity (tested with a chattering conversation).

## 5. Resumability and cost

Browser/Room presence is not execution ownership. Losing the resident socket still clears presentation and media authorization, but never cancels a turn and never kills the Harness. A reconnect performs **one bounded reconciliation**: the current execution projection is re-stated once per Task scope the Runtime owns, which is what the Room needs after it drops a replaced socket's transient projections. There is no polling, no replay of activity, and **no publication at all while a Task's state is unchanged** — a multi-hour Task costs the same handful of updates as a short one (asserted).

## 6. What is re-pinned under concurrency

Every #409/#420 invariant, unchanged: exact prepared adoption claimed on actual queue admission, `session/load` before the first turn, **zero `session/new` for an adopted scope**, no silent fresh-session fallback, session loss fail-closed. A dedicated regression runs an adopted Task and an unrelated Task **at the same time** and re-asserts all of it.

## Rolling compatibility

The Worker changes are purely additive acceptance of one new value: a legacy Runtime never sends `phase: "queued"`, so its projections validate and render exactly as before. No field becomes mandatory, no Runtime feature projection is required, and a legacy adapter that implements only the mandatory `HarnessAdapter` seam is unaffected by the new scheduler (tested).

## Tests

`agent`: 22 new tests across `harness/concurrent_turns_test.go` and `runtime/concurrent_execution_test.go`, plus a concurrent-provider mode in the scripted fake Harness (`hold_all`, per-session release, per-session permissions, opt-in chatter). Long-task coverage is expressed as policy and behavior, never as a two-minute sleep.

`app`: `queued` accepted end-to-end (contract, DO ingest, broadcast, label) and a self-contradictory queued projection rejected.

## Validation

```
cd agent
gofmt -l .                              clean
go vet ./...                            clean
go test ./... -count=1 -timeout 300s    all 12 packages ok
go test -race ./internal/runtime ./internal/harness ./internal/free4chat   ok
go test -race ./internal/daemon         FAIL: TestRuntimeExecutableCopySurvivesInstalledPathReplacement
                                        (2 DATA RACE) — PRE-EXISTING, reproduced identically on clean
                                        origin/cf-sfu before this branch; not widened
go build ./cmd/free4chat-agent          ok
bash scripts/test-bootstrap-contract.sh OK

cd app
yarn test                               106 files / 1182 tests passed
yarn type-check                         clean
yarn eslint src --no-fix                clean
yarn prettier --check .                 clean
yarn cf-build                           OpenNext build complete
git diff --check                        clean
```

## Resource measurements (real Pi through this adapter)

| State | Harness processes | RSS |
|---|---|---|
| idle (no session) | 0 | — |
| 1 session | 3 | ≈376 MB |
| 2 sessions, 2 turns concurrent | 4 | ≈398 MB |
| 4 sessions, 4 turns concurrent (probe only) | 6 | ≈622 MB |

At 4 lanes all four real Pi conversations completed at +48–49s against a 45s sleep, i.e. genuinely concurrent. 2 lanes is the product bound: the third and fourth lane cost ≈110 MB each and buy nothing the product currently needs, while the ceiling keeps one resident's local footprint predictable.

Room-side traffic shape: **one bounded request per genuine state transition** (`≈0.1 KB` JSON) and **zero while nothing changes**. No periodic publication exists to animate a spinner, so a disconnected Human costs the Room nothing while the Agent keeps working.

## Known limitations

* Codex / OpenCode / Hermes were measured concurrent with correct routing but have **not** had the isolation suite run, so they stay serial. Enabling one is a probe plus one policy value plus its provider-specific regression.
* Claude's probe could not authenticate locally; it stays UNVERIFIED and serial.
* Failure isolation is per *conversation* (verified: a killed per-session worker leaves the bridge and other sessions healthy), not per *bridge*: if the bridge process itself dies, every conversation it owned reports session loss together. That is stated rather than papered over.
* A local `free4chat-agent`/Harness restart is not durable execution. Tasks become `session lost` and require an explicit Human resume from the retained session.

NOT merged. NOT released.


madawei2699:
## Review response — two resumption correctness gaps

Both gaps are fixed on this same branch (HEAD `60be3b8`). The concurrency architecture, provider probe policy, Pi's lane bound, native-session serialization, the 6h ceiling, the opt-in idle watchdog, the scheduler, scoped interrupt, permission isolation, and the entire #420 discovery/token model are unchanged.

### 1. Supervision survives Human participant expiry

`resolveInterruptTarget` required `caller.id == Task.fromParticipantId`, so a Human who left for longer than `RECONNECT_GRACE_MS` and returned as a new participant id could send a Task follow-up and resolve the Task's permission request — both already open to any current Human — but could not Interrupt or Interrupt & send. That inconsistency is gone.

Interrupt authorization is now the same Room-shared boundary as the rest of the Task surface, and every requirement that makes the interrupt **exact** is still enforced:

* caller is a current **authenticated Human** (the shared socket guard closes unauthenticated and non-Human sockets before the handler runs);
* the canonical Human→Agent Task exists in the retained Room log;
* its canonical Agent endpoint is reachable;
* `taskRequestId` is exact — never repaired or inferred;
* `requestedTurn` matches the Room's **current transient turn** for that Task's canonical scope;
* the browser supplies no Agent, session, or native identity at all.

`fromParticipantId` remains canonical **provenance/history** and is no longer an authorization credential. The misleading "only the Human who created that Task" comments are replaced with the Room-shared contract at both call sites, and the browser interrupt control is ungated the same way — otherwise a returning Human would never *see* Interrupt on a Task they did not create.

No accounts, no Task owner/admin role, no durable participant identity, no persisted browser tokens. PRE-TASK local session selection is untouched: discovery/project/page tokens and PREPARE remain bound to the exact discovering Human (#420).

**Tests (10)**, in `roomSessionTaskSupervision.test.ts`: creator expired → returning Human sends a follow-up; returning Human interrupts the exact live turn; interrupt & send persists the replacement **before** interrupting the old turn and appends exactly one instruction plus one exact-turn control; the original creator still works. Boundaries re-asserted unchanged: unauthenticated (socket closed 4003), non-Human (closed 4003), unknown Task, stale/wrong turn, disconnected Agent, Agent with no resident socket.

### 2. Human return recovers execution state after DO hibernation

Reconciliation previously ran only when the **resident** stream reconnected, so this legal lifecycle showed a returning Human nothing at all:

```
Task Running -> Human leaves -> DO hibernates -> memory-only projections lost
-> resident socket survives, Harness keeps working -> Human returns -> nothing
```

A Human connect — and the explicit browser `resync` the client already sends on every socket open — now triggers **ONE bounded on-demand reconciliation**: a tiny private fire-and-forget `task-execution-resync` frame to each resident that advertises it. The Runtime answers with its existing `republishTaskExecutions()`, and the Room rebuilds its transient projections through the existing ingest path.

It is deliberately a **separate private family** from #409/#420 session control: no payload, no request id, no tokens, no pending-correlation state. It cannot overwrite, consume, or be answered into a live session/PREPARE correlation — proven by holding a real `task-session-list` correlation live across two resyncs and asserting the attachment is byte-for-byte unchanged and its later result still routes.

**Rolling compatibility is by advertisement, never by version string:** a new Runtime advertises `runtimeFeatures.taskExecutionReconciliation`; the Room sends nothing to a resident that does not. agent-v0.5.34 never receives an unknown private frame.

No polling, no timer, no persistence, no activity replay, no DataChannel traffic.

Also fixed along the way: `sanitizeRuntimeFeatures` dropped the whole projection unless `taskSessionContinuation` was true, which would have silently disabled reconciliation for every non-Pi launcher. Each feature is now kept independently.

**Tests (11)**: `roomSessionTaskExecutionResync.test.ts` (7) — Running and QUEUED restore after a simulated eviction (new `RoomSession` over the same storage and the same still-connected resident socket), no fabricated Running for a settled Task, explicit resync works and happens only on demand, repeated ordinary Room activity sends zero private frames, legacy Runtime receives nothing while ordinary Room/Task behavior is untouched, live #420 correlation intact and still routing. Plus 2 private-frame decode tests (`free4chat`) and 2 Runtime tests (republish without touching the session-control slot; the feature is advertised and is not launcher-gated).

### Validation on `60be3b8`

```
agent:  gofmt -l . clean · go vet ./... clean · go test ./... all 12 packages ok
        go test -race ./internal/runtime ./internal/harness ./internal/free4chat ok
        go test -race ./internal/daemon  FAIL: TestRuntimeExecutableCopySurvivesInstalledPathReplacement
              (2 DATA RACE, daemon_test.go:1087/1091) — PRE-EXISTING, reproduced identically
              on clean origin/cf-sfu; not widened
app:    yarn test 108 files / 1201 tests passed (+19)
        yarn type-check / eslint / prettier clean · yarn cf-build OpenNext build complete
        git diff --check clean
```

No release artifacts touched: `doctor.Version`, `app/public/agent.md`, `.github/`, and every release script are byte-identical to `cf-sfu`.

### The #421 acceptance statement, now literally true

| Lifecycle | Before | Now |
|---|---|---|
| Human short network reconnect | supervision + state OK | unchanged |
| Human leaves > 30s, original participant expires, **fresh Human returns** | could follow up and resolve permissions, **could not interrupt**; state recovered only if the resident also reconnected | current authenticated Human recovers truthful state on connect and can follow up, interrupt, and interrupt & send the same canonical Task |
| Resident Agent transport reconnect | Runtime republishes execution truth | unchanged |
| DO hibernation with resident socket surviving | **nothing recovered**; Human saw no Running projection | **one bounded on-demand reconciliation** on Human connect/resync; truthful Running/QUEUED restored, and no phantom Running for a settled Task |

NOT merged. NOT released.

madawei2699:
## Real serial Harness regression (pre-merge smoke)

Run against this exact HEAD (`60be3b8`) before merge, on the real local installations. **No code changes were required**, so this section is a record, not a patch: the probe was temporary and is not committed.

### What was actually under test

The shared `ACPAdapter` and `ResidentRuntime` were rewritten in this PR (per-session `activeACPTurn`, `activeTurns` lane set, scoped cancel, long-turn ceiling + opt-in idle watchdog, the queue scheduler). Pi has the full #421 concurrency/isolation probe. The other three Harnesses stay **SERIAL / one lane**, so what had to be re-proven for them is that the new shared machinery still behaves exactly like the old serial path.

**Real** here means: the launcher from the shipped registry, the real ACP bridge and provider CLI, the real `ACPAdapter` with production defaults (`TurnTimeoutMs` unset → the shipped 6-hour ceiling, so the *new* policy is what is in play), the real `ResidentRuntime` scheduler, the real exact-turn interrupt path, and the real Room-mediated permission bridge (`RequestPermission` → projected Room card → decision on the resident event stream). The Room was stood in for in-process, because a real Room needs a deployed Worker; the Harness, bridge, adapter, scheduler, timeout policy, interrupt path, and permission path were all production code.

Workload: a genuine read-only analysis of a real 138-file Go tree copied into a disposable workspace — the Harness must recursively read every file and produce a one-line-per-file report. Nothing in any user workspace was touched, and the durability of the test is bounded by the provider finishing real work, not by a sleep. If a provider happened to finish the real analysis quickly, the long-task phase is retried (max 3) rather than weakened, because "the model finished early" is a property of the workload, not of the product.

### Results

| Harness / version | launcher (pinned) | policy | Task A > 2 min | Task B queued | no accidental concurrency | Interrupt A isolated | B runs after capacity released | A follow-up continuity | permission | unexpected process / session behavior | verdict |
|---|---|---|---|---|---|---|---|---|---|---|---|
| **Codex** `codex-cli 0.147.0` | `npx -y @agentclientprotocol/codex-acp@1.6.2` | `serial`, `Lanes()=1` | **2m29s**, 0 timeout failures, 0 cancels | **yes**, `phase=queued`, A untouched | **yes**, peak = 1 | **yes**, exact turn stopped at 2m33s | **yes**, `B-DONE` | **yes**, "Subject: auditing and summarizing the workspace's Go files. SERIALSMOKE-FOLLOWUP" | n/a (read-only mode) | none | **PASS** (attempt 1/3) |
| **OpenCode** `1.18.18` | `opencode acp --pure` | `serial`, `Lanes()=1` | **2m17s**, 0 timeout failures, 0 cancels | **yes**, `phase=queued`, A untouched | **yes**, peak = 1 | **yes**, exact turn stopped at 2m20s | **yes**, `B-DONE` | **yes**, retained the original instruction incl. its `SERIALSMOKE-A` token | **yes — 9 real Room approval cards** with project `permission: bash/edit/webfetch = ask`; attributed 3 → `req-A`, 6 → `req-B`, i.e. each to its OWN Task, never a foreign one; every card offered 3 options and a bounded `expiresInMs=300000`; all answered allow-once | none | **PASS** (attempt 1/3) |
| **Hermes** `Hermes Agent v0.21.3 (2026.9.14)` | `hermes acp` | `serial`, `Lanes()=1` | **2m19s**, 0 timeout failures, 0 cancels | **yes**, `phase=queued`, A untouched | **yes**, peak = 1 | **yes**, exact turn stopped at 2m22s | **yes**, `B-DONE` | **yes**, "Subject of this conversation: a real read-through of the Free4Chat Go Agent Runtime source tree — all 138 `.go` files … SERIALSMOKE-FOLLOWUP" | n/a | none (wrote its report into the disposable workspace only) | **PASS** (attempt 3/3; attempts 1–2 finished the real analysis early, which is model variance, not a product result) |
| Pi | `npx -y pi-acp@0.0.33` | `cross-session`, `Lanes()=2` | covered by the full #421 concurrency/isolation probe; not repeated | | | | | | | | — |
| Claude | `npx -y @agentclientprotocol/claude-agent-acp@0.70.0` | `serial`, `Lanes()=1` | deferred to a separate machine before the combined release | | | | | | | | — |

Launcher policy re-verified at this HEAD (not flipped anywhere): `hermes` / `opencode` / `codex` / `claude` / custom → `concurrency=serial`, `Lanes()=1`; only `pi` → `cross-session`, `Lanes()=2`.

### What this proves

```
policy = serial / lanes = 1
under the NEW  activeACPTurn · activeTurns · scoped cancel
              long-turn ceiling · bounded queue scheduler
still behaves exactly like the old serial path
```

Specifically: the old 120-second product cap is gone for every Harness (real turns of 2m17s–2m29s with zero `failureClass:timeout` and zero dispatched cancels); a second independent Task truthfully reports `Queued` instead of appearing hung and never reaches the Harness while the first owns the lane (observed peak concurrency = 1 for all three); the scoped interrupt stops exactly the named Task's turn and leaves the other untouched; the queued Task then obtains the single lane and completes; the retained Task conversation survives the interrupt and answers a later follow-up on the same conversation; and OpenCode's real interactive approval policy routes each card to the exact Task that asked for it, with an answered approval resolving only its own Task.

No regression was found, so **no code was changed** for this smoke.

CI on `60be3b8` remains green (App Tests, Lint & Type Check, gofmt/vet/test/build, build matrix ×4, Room App host matrix, Distribution/cleanup contract, Format & lint autofix; Build & Deploy and Publish GitHub Release correctly skipped).

NOT merged. NOT released.

Issue #429, opened 2026-09-20
bug(agent): preserve long Codex command approval options in Room projection


madawei2699:
## Summary

Codex command approvals can be dropped before a Room request is created when a bridge supplies a native option label longer than Free4Chat's current Room presentation limit.

## Reproduction

Using `@agentclientprotocol/codex-acp@1.12.0` with its resolved `@openai/codex@0.154.0`, `INITIAL_AGENT_MODE=read-only`, `approval_policy=on-request`, and `approvals_reviewer=user`, an out-of-workspace `touch` produced an app-server `item/commandExecution/requestApproval`. The bridge forwarded it as ACP `session/request_permission` with a valid title and permission description.

One offered native option, `accept_execpolicy_amendment`, had a 175-character display name. Runtime validation caps option names at 160 characters, so `projectRoomPermission` returned `invalid Harness permission option name` before calling the Room permission client. No Room request/card was created and the requested command did not execute.

## Expected

Preserve the exact native option ID while making a bounded Room-safe display label, or otherwise report a visible unsupported approval state. A long provider label should not silently cancel an otherwise valid approval request.

Add a regression for a native permission option name over the current limit.

Related verification: #426.

madawei2699:
Fix is in #430 (`fix(agent): bound long native approval labels instead of dropping the request`).

Native identity and Room presentation are now separate: the opaque `OptionID` stays exact, and an over-budget human-facing label becomes a bounded Room label inside the **same** 160-unit presentation budget, with a visible ellipsis and surrogate-safe truncation. No limit was raised, no provider branch was added, no option was invented or dropped.

Regression coverage in `agent/internal/runtime/permissions_projection_test.go`: short label unchanged byte-for-byte; long Codex-shaped `accept_execpolicy_amendment` label projected with the exact native ID still in `offered`; Room decision round trip carrying the exact original ID (a display label is never accepted as an identity); Unicode/emoji surrogate boundary; and the unchanged fail-closed cases (empty/whitespace labels, empty/whitespace-mutated/oversized/duplicate IDs).

Leaving this issue open until #430 is merged.

madawei2699:
Fixed by merged PR #430 (merge commit `ce6a62b23ca57226c26d79b90ced9b3174700747`).

The fix keeps native `OptionID` exact and bounds only the Room-facing display label within the existing 160 UTF-16-unit presentation budget, with surrogate-safe truncation and focused regression coverage.

[...]

Issue #427, opened 2026-09-20
architecture(agent): establish a Free4Chat-owned Harness semantic compatibility layer


madawei2699:
## Why this exists

Recent Remote Task dogfood exposed a repeated integration risk:

Free4Chat currently relies on multiple Harness/ACP implementations with materially different semantics:

- Hermes — native ACP
- OpenCode — native ACP
- Codex — `@agentclientprotocol/codex-acp`
- Claude — `@agentclientprotocol/claude-agent-acp`
- Pi — `pi-acp`
- future built-ins may add more implementations

ACP is a useful transport/protocol boundary, but **ACP support alone is not a sufficient product contract for Free4Chat**.

Two concrete production-grade examples now prove that:

### Pi session lifecycle

Free4Chat needs:

```text
load another retained session
→ must not destroy an unrelated active Task
```

But `pi-acp@0.0.33` uses `closeAllExcept()` from `session/new` / `session/load`, which can kill another active per-session Pi worker and leave its `session/prompt` unresolved.

The Free4Chat Runtime/lane design was not the root cause. A production-shaped probe against upstream PR `svkozak/pi-acp#131` verified that preserving concurrent session workers restores the desired behavior:

```text
load A
→ prompt A active
→ load B
→ A continues
→ A/B make real concurrent progress
```

Track the user-facing Task consequence in #421.

### Codex approval behavior

Free4Chat needs:

```text
provider requests an exact local approval
→ Free4Chat can present it remotely
→ exact decision returns to the same turn
```

But compatibility depends on both Codex app-server behavior and the exact `codex-acp` version/policy. #297's historical result was based on an older bridge; current re-verification is tracked in #426.

These are not isolated bugs. They demonstrate a repeated boundary problem:

> Runtime code should depend on Free4Chat-owned semantics, not on assumptions about how a particular ACP implementation interprets the protocol.

Refs #297, #409, #421, #426.

---

## Architectural principle

Do **not** replace ACP.

Instead:

```text
Free4Chat Runtime
        ↓
Free4Chat Harness semantic contract
        ↓
provider-specific compatibility adapter
        ↓
ACP transport / native Harness protocol / bridge
        ↓
Harness
```

ACP remains the primary transport where appropriate.

Free4Chat owns the product-level semantics above it.

The compatibility layer absorbs:

- provider/bridge quirks;
- version-specific workarounds;
- capability normalization;
- safe fallbacks;
- upstream compatibility fixes;
[...]

Issue #426, opened 2026-09-20
compat(agent): re-verify Codex Room approvals on current codex-acp


madawei2699:
## Why

Free4Chat #297 classified Codex Room approval as an upstream limitation based on older `codex-acp` versions.

Free4Chat currently pins `@agentclientprotocol/codex-acp@1.6.2`. Upstream changed materially after that: codex-acp 1.7.0 added ACP v1 permission presentation (agentclientprotocol/codex-acp#405). Current latest is 1.12.0, and current main contains explicit mappings for `item/commandExecution/requestApproval`, `item/fileChange/requestApproval`, `item/permissions/requestApproval` → ACP `session/request_permission`.

The old Free4Chat compatibility result therefore has to be re-verified against current upstream.

## Remaining upstream uncertainty

Permission presentation support does NOT imply every Codex approval path is correct. Still-open upstream evidence:

- openai/codex#21982 — `sandbox_permissions` / `require_escalated` may fail to surface `requestApproval`; issue remains OPEN, similar symptom reported on newer Codex versions.
- agentclientprotocol/codex-acp#310 — bridge mode presets / config policy limitations.
- agentclientprotocol/codex-acp#401 — some MCP execution can occur before permission mediation.

## Goal

Determine current real behavior, not design a new Free4Chat permission system. Classify the current Codex bridge as one of:

- `ROOM APPROVAL VERIFIED`
- `PARTIAL`
- `UPSTREAM APP-SERVER LIMITATION`
- `UPSTREAM BRIDGE LIMITATION`

## Acceptance

- probe current `codex-acp` 1.12.0 against the current Free4Chat pin 1.6.2 in a disposable workspace;
- choose the narrowest mode/policy that is supposed to produce manual approval (do not assume old INITIAL_AGENT_MODE semantics; do not permanently widen Free4Chat's read-only policy — an isolated alternate mode/config is acceptable for the probe);
- trigger real command/file approval and trace the exact layer:
  Codex app-server `requestApproval`
  → `codex-acp` `session/request_permission`
  → Free4Chat Room approval card
  → exact native option
  → same turn continues (allow) / action prevented (deny);
- if PASS: run the Codex serial regression on 1.12.0 (ordinary Task, >2 min turn, Task B queued, interrupt, follow-up, session behavior, permission allow/reject) and only then decide whether Free4Chat should bump its launcher pin — Codex stays serial, no concurrency, `TaskSessionContinuation` unchanged;
[...]

PR #430, opened 2026-09-20
fix(task): stabilize Remote Task production dogfood UX (#421)

madawei2699:
## Why

v0.5.35 production dogfood proved the long-lived Remote Task model works: a
real engineering Task ran for 30+ minutes, was started and supervised from a
phone, survived the Human leaving the Room and returning from another device,
retained its Pi session, and handled Interrupt & send correctly.

It also exposed a small, concrete set of **user-facing** stabilization gaps.
This PR fixes those and nothing else.

Refs #421
Refs #409
Refs #429

Not merged. Not released. **#421 stays OPEN.** #426 is not closed.

---

### 1. Task execution is the control authority (#421 Fix C)

`TaskExecutionProjection` and `AgentActivity` were two independent transient
Room stores, fed by two independent Runtime requests. Room interrupt
authorization read the **presentation-only** Activity map, while the #421
reconciliation path restores **execution truth only** — so a hibernated Room
could show a Task as Running and offer Interrupt, then refuse the click with
`task_turn_not_active`.

* `TaskExecutionProjection` is now the single authoritative Task execution and
  control state; `AgentActivity` supplies the coarse verb only.
* The exact-turn requirement is **not weakened**: the requested turn must equal
  the authoritative current turn of that exact canonical Task.
* The Browser now receives the execution projection as a **live** frame
  (`taskExecution`) instead of only on a full Room `state`, so reconciliation
  actually reaches a returning Human. This was a real second defect: the Room
  broadcast the projection and the client silently ignored it.
* The Interrupt control is offered from the authoritative projection, so a
  missing or stale `AgentActivity` can no longer make a genuinely running Task
  uncontrollable — and a stale activity can no longer invent authority.

### 2. Reconnect-safe interrupt

Task running → Human disconnects → Room DO may hibernate → Task continues
locally → Human returns → the existing #422 `task-execution-resync` re-states
the exact current turn → **Interrupt succeeds**. No polling, no timer, no
persisted Task DB state.

### 3. Interrupt & send race (#421 Fix D)

Previously the exact-turn check **gated the instruction itself**, so a turn that
settled a millisecond before the command landed lost the Human's replacement
entirely and surfaced a raw protocol code.

Now: Task/agent authorization stays fail-closed, the instruction is appended
**exactly once**, and the interrupt is dispatched **only** while that exact turn
is still authoritative. A stale request can never cancel a later turn, and the
replacement continues as a normal queued Task instruction.

### 4. Raw control errors (#421 Fix G)

Benign control races are reported as a bounded `task-control-notice` and
rendered as **transient local feedback** next to the Task controls, never as the
sticky Room-wide banner:

* standalone stale Interrupt → *"This turn has already finished."*
* stale Interrupt & send → *"Current turn already finished; instruction sent."*

Genuine failures (unknown Task, unreachable Runtime, malformed protocol input,
permission failure) keep the ordinary error path unchanged.

### 5. Task artifact correlation

`cb8cd5f` (already on this branch) is **reviewed and retained**: the Harness turn
input carries the exact canonical Task requestId, and `RenderUntrustedRoomTurn`
states the concrete `attach --file <path> --task-request-id <id>` command. A Room
turn keeps an empty id, so an unscoped `attach` remains a Room artifact and the
Runtime never infers a global "current Task". Two concurrent Tasks stay
explicitly correlated by construction.

### 6. #429 — long native approval labels

A real `codex-acp 1.12.0` command approval offered `accept_execpolicy_amendment`
with a ~175 UTF-16-unit display label. The Room presentation budget is 160, and
`projectRoomPermission` rejected the **entire request** on that budget — no Room
card, nothing for the Human to approve, and the command never ran.

Native identity and Room presentation are now separate: the opaque `OptionID`
stays exact, and an over-budget *human-facing label* becomes a bounded Room label
inside the **same** budget with a visible ellipsis. Truncation is surrogate-safe.
No limit was raised, no provider branch was added, no option was dropped or
invented, and empty/whitespace-only labels, oversized IDs, whitespace-mutated IDs
and duplicate IDs all still fail closed.

---

### Large Start Task brief (#421)

The Start Task modal bounded its instruction at the canonical 1200-character
collaboration summary, so a real handoff had to be created as an empty Task and
pasted afterwards as a follow-up. The product no longer requires that trick.

One shared big-paste primitive now backs both composers. Above the existing
threshold a pasted document becomes a Task-correlated `text/markdown`
attachment, byte for byte, and the canonical Task event **references it
explicitly** — so the brief is inside the Task before its first wake and the
Agent's first turn already has it. One canonical Task, one addressed wake, one
attachment, no duplicate Room artifact.

The flow fails closed: if the brief cannot be staged, **no Task is created** and
the modal keeps the Human's text. A paste beyond the bounded attachment store is
refused visibly instead of being truncated.

---

## Mobile

Phone dogfood proved the Room was usable but badly composed: the participant /
Stage panel is a `flex-1` sibling of the chat panel and took ~43% of the
viewport *before* any interaction.

Below `md` the participant/Stage panel is now secondary — hidden by default so
the active Room/Task interaction owns the viewport — with one compact header
line (`N participants · Pi · Using tools…`) and a single `⋯` overflow that
reveals the same one mounted panel as a full-height sheet. The header keeps only
the interaction title, that signal and the overflow; Copy link / Invite Agent /
Live Transcript / mic / Leave join the overflow.

The sheet is **geometry, not a copy**: it wraps the one already-mounted panel, so
participant media nodes, Room App iframes and MessagePorts are never remounted
and no control is duplicated. Every `md:`/`lg:` token, the resize handle, the
split width and the fullscreen guards are unchanged.

---

## Session Picker

Production dogfood showed long raw project paths made the bounded project
control awkward. The display path is now compacted while preserving the
distinguishing tail; search still matches the **full** label and the untouched
path stays available as the element `title`. The Runtime label is
presentation-only and never sent back, so #420's privacy boundary is unchanged.

---

## Measured browser evidence

Real local Worker + real `RoomSession` Durable Object
(`app/e2e/room/run-local-worker.mjs`), Chromium, 390×844 and 1440×900:

| metric (390×844) | before (`cb8cd5f`) | after |
| --- | --- | --- |
| header height | 122 px | **69 px** |
| participant panel | 362 px tall, always visible | not rendered until the overflow opens it |
| active interaction top | 529 px | **114 px** |
| active interaction height | 316 px (37%) | **730 px (86.5%)** |
| composer bottom | 832 px | 832 px (in view) |
| horizontal overflow | 0 px | **0 px** |
| secondary toolbar in permanent chrome | yes | no (overflow only) |

Desktop geometry is **identical** before/after: header 53 px, participant panel
720×847, chat panel 716 px wide, interaction top 98 px, height 802 px, composer
bottom 888 px.

Before/after screenshots (desktop, 390×844, sheet open, Task view) were captured
with a Playwright script against that same harness.

---

## Targeted end-to-end dogfood

Real local Worker + real `RoomSession` DO + the **real Go Agent Runtime**
(resident, out-of-process) + a real phone-viewport browser, driven by the repo's
scripted ACP Harness holding every turn until a genuine `session/cancel`:

| step | result |
| --- | --- |
| paste a 4 246-character brief into Start Task | one canonical Task, derived label, brief as Task context |
| brief in Room scope? | **no** (no Room artifact) |
| Task running | `task-interrupt` offered, signal `2 participants · Pi · Working…` |
| **full disconnect → new browser → rejoin** | Interrupt still offered while Running |
| click Interrupt | turn settled `Interrupted`; Runtime log `task_interrupt_requested` → `task_interrupt_settled` |
| raw error banner | none |
| second Task → Interrupt & send | replacement stored **exactly once**, running turn cancelled, **next turn started** |

The same run is what caught the Worker header allow-list bug below.

### Bug the dogfood caught

`app/src/room/server.ts` is a header **allow-list**, not a transparent proxy, so
the new pre-Task marker was silently dropped: the Room correctly refused an
unknown Task and the modal reported the brief could not be attached. The marker
now crosses as exactly one canonical token, with a Worker-level regression test.

---

## Validation

Agent (`cd agent`):

* `gofmt -l .` — clean
* `go vet ./...` — clean
* `go test ./... -count=1 -timeout 300s` — **all packages ok**
* `go test -race ./internal/runtime ./internal/harness ./internal/free4chat -count=1` — **ok**
* `go build ./cmd/free4chat-agent` — ok
* `bash scripts/test-bootstrap-contract.sh` — **OK** (all PASS)

App (`cd app`):

* `yarn test` — **114 files / 1249 tests passed**
* `yarn type-check` — clean
* `yarn eslint src --no-fix` — clean
* `yarn prettier --check .` — clean
* `yarn cf-build` — ok

Root: `git diff --check` — clean.

New deterministic coverage:

* Room: interrupt-authority matrix A–H
  (`roomSessionTaskControlAuthority.test.ts`), execution/resync, supervision and
  interrupt suites updated to the authoritative model; Start Task brief suite;
  Worker header transport suite.
* App: shared big-paste primitive; Start Task brief UI (inline vs attachment,
  brief-only start, short instruction + context, fail-closed, over-limit).
* Agent: `permissions_projection_test.go` — short label unchanged, long
  Codex-shaped label bounded with the exact native ID preserved, decision round
  trip by exact ID, Unicode/surrogate boundary, and the fail-closed
  identity/label cases.

---

## External blockers — not fixed here

* **Pi cross-session concurrency** — the production-shaped failure is upstream
  `pi-acp@0.0.33` lifecycle behaviour; independently verified fixed by
  [svkozak/pi-acp#131](https://github.com/svkozak/pi-acp/pull/131). No
  serialization workaround, no process pool, **no lane-policy change**, no fork.
  Room=1 + Tasks=2, `MaxConcurrent*` and the Pi launcher distribution are
  untouched.
* **Codex approval re-verification** — tracked in #426. The launcher stays
  pinned at `@agentclientprotocol/codex-acp@1.6.2`; Codex execution policy stays
  SERIAL and TaskSessionContinuation stays disabled. #429 fixes the Free4Chat
  projection bug only; #426 still needs a real compatibility acceptance against
  codex-acp 1.12.0 (live browser Room card, Allow executes in the same turn,
  Reject does not execute) plus the broader serial regression before any pin
  bump. #426 is **not** closed.
* **Harness semantic architecture** — #427, implemented separately. This PR
  contains no provider abstraction and no harness refactor.
* **Remote Operator Approval without a provider-native blocking seam** —
  recorded as an upstream/provider limitation. No fake permission system, no
  command parser, no invented fallback option. Ordinary Room chat stays
  untrusted.

### Architecture overlap with #427 — reported as required

This branch changes exactly one file under `agent/internal/harness`
(`prompt.go`, from the already-present `cb8cd5f`), and #427 touches
`launchers.go` plus new `provider*.go` / `contract.go` / `capabilities.go`
files. The overlap is therefore **limited to the directory**, not to a hunk, and
the behavior fix is the smallest possible: the rendered prompt states the
concrete Task requestId. Product correctness takes priority over structural
cleanup, and this PR does not attempt #427.


madawei2699:
Review follow-up: the first Room App Host Compatibility run failed only on `webkit-phone` during the shared `enterLocalRoom` fixture.

The browser had already joined successfully (`1 participant · … · Room · No messages yet`). The fixture still required the self participant card (`Alice`) to be **visible**, but #430 intentionally keeps the single Stage/participant panel mounted while display-hiding it on phones until the overflow opens.

Commit `76054c58` updates the E2E joined-state invariant from **visible** to **attached**. This is test-only; no product code or Room App lifecycle code changed. The self card remains mounted, which is exactly the new mobile composition contract.

Fresh CI is running from `76054c58`.

madawei2699:
Second Room App host CI follow-up: after the joined-state fixture was corrected, the phone profile reached the next stale assumption in the spec itself: line 340 still asserted the Stage must be visible immediately after join.

That contradicts #430's intentional phone contract (Stage is one persistent mount but hidden until `⋯` opens it). Commit `8806daf6` updates the compatibility path to exercise the real product flow:

phone → assert Stage hidden → click `⋯` → assert the same Stage visible → continue the existing App open/fullscreen/exit/hide/reopen contract.

Desktop/tablet two-pane behavior keeps the existing immediate Stage-visible assertion. Test-only change; no product code changed. Fresh CI is running.

PR #433, opened 2026-09-20
docs(task): document the shipped Task semantics (#421, #409)

madawei2699:
Documentation-only follow-up to the merged v0.5.35 Task work (#430). It is
**separate from** the Codex compatibility/pin PR (#432) and touches no source
code.

Refs #421
Refs #409

## What changed

**`docs/en/guides/tasks-and-live-views.md`** — five new Human-facing sections:

* **Starting a Task** — a Task starts a new Harness session by default; an
  existing local session can be chosen only where continuation is *verified*
  (not merely advertised). Discovery is local/private and lazy: nothing is read
  from the machine just because Start Task was opened, and the local session
  list is requested only when the continuation path is explicitly opened. A
  failed continuation start creates no Task and never silently falls back to a
  new session.
* **Large initial briefs** — a substantial brief can be pasted straight into
  Start Task. When it is too large for the inline instruction it is preserved as
  **bounded Task-scoped context**, attached to the canonical Task *before* its
  first Agent turn, never summarised, rewritten, or silently truncated. If that
  context cannot be staged, the Task is **not** started and the text stays in
  the dialog. The existing public 768 KB text-like bound is kept; the current
  inline UX threshold is deliberately **not** documented as a contract.
* **Long-running and resumable supervision** — the browser is not the owner of
  local execution; closing/leaving does not by itself cancel a running local
  Task, and returning reconciles bounded coarse execution state and restores the
  Task controls. States are described truthfully (`Starting`/`Working`,
  `Running`, `Queued`, `Interrupted`, `Completed`/`Failed`, `Session lost`), and
  a pending approval is described as a card rather than invented as a status.
  Human/browser disconnect is explicitly distinguished from local
  Runtime/Harness process death, with no durable-execution promise across a
  local process, daemon, or machine shutdown. Room expiry is unchanged.
* **Controlling a running Task** — Interrupt stops the exact active Task turn;
  Interrupt & send keeps the replacement instruction and interrupts the current
  turn while it is still active; if that turn already finished, the instruction
  still becomes the next Task instruction and nothing later is cancelled. All at
  product level — no protocol names or error codes.
* **Execution capacity** — one native Harness conversation is always serialized;
  independent Tasks may make bounded concurrent progress **only** where that
  Harness has been verified safe; full capacity shows as `Queued`. No claim that
  every built-in Harness supports concurrency.

**`docs/en/concepts/runtime-harness.md`** — one new section adding the ownership
distinction:

```text
browser/Room transport  !=  local execution ownership
```

with the narrow three-case explanation (disconnect → local may keep executing;
reconnect → the Runtime reconciles bounded shared execution state; local process
death → continuous execution is not guaranteed) and the explicit product
boundary: Free4Chat is **not** a cloud job runner, permanent workspace, or
durable execution service.

**`app/public/llms-full.txt`** — regenerated through the repository's normal
docs generation path, not edited by hand.

## Deliberately excluded

Mobile layout implementation details; internal execution-projection terminology;
provider-specific bridge bugs; permission-label internals; ACP package/version
details; speculative future capabilities. Nothing here promises provider-specific
behavior that is not generally true.

## Checks

* `node app/scripts/generate-docs-assets.mjs` then `--check` — **docs assets OK**
  (`llms.txt`, `llms-full.txt`, `sitemap.xml`)
* `npx vitest run src/__tests__/docs src/__tests__/discovery/crawl-surface.test.ts`
  (the CI docs contract suite) — **41 passed**
* `npx prettier --check .` (from `app/`) — clean
* `git diff --check` — clean
* Change set is exactly 3 files: 2 Markdown pages + the generated artifact. No
  source, test, or configuration change.


madawei2699:
Merge-order follow-up: #432 (Codex 1.12.0 compatibility) landed first as `c6ec2ab812167f1e9299ad0e3ff3f6c950de9a6d`.

This docs branch is now merged forward onto that base at `62505c137dddd495b96250af6030270f0ff26511`.

The regenerated public corpus now contains BOTH:
- #432's current Codex approval documentation; and
- this PR's Task/runtime ownership documentation.

The source docs remain exactly the same two Task pages from this PR; the third changed file is still the generated `app/public/llms-full.txt`.

Fresh CI is running against the combined state before merge.

Issue #441, opened 2026-09-20
enhance(task): allow explicit Agent handoff after the current executor disconnects


madawei2699:
## Why

Human production dogfood on v0.5.36 exposed a Task ownership semantic gap.

Real scenario:

- Task was started with Pi on machine A;
- machine A crashed / Pi left the Room;
- Claude remained connected from machine B;
- Human opened the existing Pi Task and explicitly addressed Claude to review/continue the work;
- Room returned `task_target_not_in_room` before Claude could join the Task.

This is not the desired product model. A canonical Task is a Room collaboration object, not the lifetime of one Agent process or one native Harness session.

## Current code path

`app/src/do/taskScope.ts` already supports extending Task participation when a Human explicitly targets another connected Agent:

- `resolveHumanTaskTargets(...)` accepts any explicitly targeted connected Agent;
- `buildTaskProjectionIndex(...)` adds validated Task text targets to `participatingAgentIds`;
- historical Task projection then allows that newly participating Agent to inspect retained Task context.

But `resolveTaskRequest(...)` runs first and currently fails when ZERO already-participating Agents are connected:

`if (agentParticipantIds.length === 0) return { ok: false, error: "task_target_not_in_room" }`

So the Human never gets the chance to explicitly add Claude after Pi disconnects.

## Product semantics

Keep these identities separate:

`Task identity != Agent process != native Harness session`

When the current executor disappears:

- the Task remains canonical and reviewable while its retained request/context still exists;
- the old executor/native session may truthfully become unavailable / Session lost;
- Human may explicitly target a DIFFERENT connected Agent in the same Task;
- that explicit Human action admits the new Agent into Task participation;
- the new Agent receives the bounded retained Task context/artifacts it is authorized to see;
- the new Agent starts/uses its OWN Harness session unless its own provider semantics say otherwise.

Do NOT pretend Claude inherited Pi's native conversation/session.

Do NOT automatically fail over to another Agent without Human intent.

## Desired flow

`Task T -> Pi running -> Pi disconnects -> Task T: no active executor / Session lost -> Human explicitly @Claude -> Claude admitted to Task T -> Claude receives retained Task context -> Claude reviews/continues in its own Harness session`

Once Claude has been explicitly admitted, future Task messages may use Claude as a participating endpoint according to the existing Task participation rules.

## Security / authority

Fail closed on implicit routing, but not on explicit Human reassignment.

[...]

Issue #436, opened 2026-09-20
bug(task): repeated interrupt/recovery can desynchronize Task execution


madawei2699:
## Why

Human production dogfood on `agent-v0.5.36` found a correctness gap after repeated Task interruption.

This is distinct from the known Pi `pi-acp@0.0.33` retained-session concurrency bug.

Observed provider results:

- Claude: interrupt → recover → continue passed in one Task.
- Codex 1.12.0: interrupt → recover → continue passed.
- Hermes: first interrupt/recovery worked; on a later interrupt the native Harness session was actually interrupted, but Free4Chat still showed the Task as Working and subsequent interaction in that Task stopped progressing.
- Pi: interrupt reached the native Pi session at least once, but subsequent addressed input became inconsistent. The Room showed e.g. `Running · 4 queued`; sending an interrupt/replacement temporarily moved the count 4 → 5 → 4, and one addressed message appeared in Pi's native session while a later one did not. No clear Room error surfaced.

One captured Pi native-session prompt after interrupt contained the exact canonical Task context and the addressed Human message, proving at least one replacement crossed the Runtime/Harness boundary:

```text
Current Task requestId: cc3a99ee-5869-4e8d-9861-c777b53da7d2
...
Pellen (human) [addressed]: 你注意本地有 repo 啊，你能找到的到吗？
```

The problematic shape is therefore not simply "cancel never reaches the Harness".

## Product invariant

For one Task:

```text
interrupt exact active turn
→ native turn settles/cancels
→ authoritative Task execution state reconciles
→ queued replacement/follow-up is admitted exactly once
→ Task either keeps progressing or reaches a truthful terminal/session-lost state
```

Never:

- leave the Task visually `Working` after the underlying turn is gone;
- silently strand queued Human input;
- consume/drop one replacement while keeping later queue counts misleading;
- require a browser refresh to discover that the Task is dead.

## Investigation

Trace the exact sequence across:

- Room canonical Task input append;
- TaskExecutionProjection;
- Runtime queue admission/drain;
- scoped turn ownership;
- ACP `session/cancel` settlement;
- next-turn wake;
- provider process/session state.

Reproduce at minimum:

1. Task turn A running.
2. Interrupt or Interrupt & send.
3. replacement/follow-up executes.
4. second long turn runs.
5. interrupt a second time.
6. another follow-up executes.

Run separately against:

- Hermes native ACP;
- Claude;
- Codex 1.12.0;
[...]

PR #446, opened 2026-09-20
fix(task): settle execution state after turn and executor termination

madawei2699:
## Fixed

- A disconnected Agent can no longer resurrect a Task execution projection through its still-valid participant handle.
- After the original executor has departed, a Human-explicitly admitted replacement Agent can publish the authoritative execution projection for its own Runtime turn.
- The Task UI renders the Room-accepted replacement execution projection and its exact interrupt control instead of treating the initial Agent as the permanent executor.

## Regression coverage

- running -> normal return -> no current turn in the Room-facing projection;
- running -> exact interrupt settlement -> no current turn plus interrupted outcome;
- executor departure clears active ownership and blocks stale re-publication;
- A departure -> Human explicitly targets B -> B can own and settle its own execution projection;
- existing explicit-target and unrelated-Agent routing tests remain covered.

## Still provider-dependent

This does not claim that a native provider cancel which never settles has settled. Real repeated-interrupt provider behavior remains tracked in #436, and Pi remains unchanged. The exact post-departure real-provider handoff smoke must run after this app/DO change is deployed; this PR does not deploy or release.

Refs #436 #441


madawei2699:
Final review found one small merge blocker in the browser projection selection.

`RoomContent.tsx` currently resolves the active Task execution as:

1. projection for the Task's original `targetParticipantId`;
2. otherwise the first projection with the same `taskRequestId`.

But Room execution projections are stored per `agentParticipantId + taskRequestId`, and #441 already permits multiple Agents to participate in one canonical Task. Therefore A and B can both have retained transient projections for the same Task while both are still connected. In that case an explicitly selected replacement/secondary Agent B may be the current executor, but the UI will still prefer A's projection solely because A created the Task.

Please fix this before merge. Do NOT add a browser heuristic like "prefer the newest array item" unless the ordering is itself authoritative.

The UI needs one unambiguous authoritative Task execution selection rule derived from server-accepted execution truth. Prefer the smallest option that fits the existing model, for example ensuring the Room exposes at most one current/active execution projection per canonical Task (while preserving terminal outcome as needed), or otherwise carrying the exact current executor identity explicitly.

Add regression coverage for:

- Task initially targets A;
- A remains connected;
- Human explicitly targets/adopts B in the same Task;
- B publishes Running for its own turn;
- A has an older/terminal projection for the same Task;
- UI must show B as the active executor and Interrupt must target B's exact turn;
- unrelated participating Agent projections must not win selection.

The executor-loss path already covered by this PR should remain unchanged.

Once this ambiguity is removed and CI is green, #446 should be mergeable.

madawei2699:
One final semantic blocker before merge: the latest fix makes `transientTaskExecutions` unique by `taskRequestId`, but the existing Task model explicitly supports multiple Human-selected targets (`resolveHumanTaskTargets()` accepts a bounded array of connected Agents). Those Agents are separate residents and can legitimately have overlapping real executions inside the same canonical Task.

Therefore collapsing server truth to ONE projection per Task can silently lose a live execution:

```
Task T
Human explicitly targets A + B
A publishes Running(turn 10)
B publishes Running(turn 20)
=> current PR overwrites A with B even though A may still be running
```

That would make Room execution truth less accurate and can misroute/under-report control state.

Please preserve the existing per-Agent execution projections on the server. Fix the original handoff bug without deleting multi-Agent truth.

A narrow safe shape is:

1. keep transient execution identity as `(agentParticipantId, taskRequestId)`;
2. departed Agent projections are still cleared and disconnected handles still rejected;
3. newly explicitly admitted B may publish its own projection;
4. derive/select a SINGLE control executor only when it is unambiguous from authoritative state:
   - executor-loss case: A projection is gone, B is the sole active execution -> B;
   - A terminal + B Running -> B is the sole active execution -> B;
   - exactly one active projection (`currentTurnSequence`) -> that Agent owns Interrupt / Interrupt & Send;
   - >1 active executions -> do NOT pick by initial target, array order, timestamp, or last writer. Preserve all truth and fail closed for any control that requires one exact executor unless there is already an explicit server-side ownership primitive that resolves it.
5. terminal projections must never outrank a unique active projection.

If the existing protocol/UI needs a tiny aggregate helper for the selected exact active execution, add that narrowly; do not introduce a scheduler or redesign Task UI.

Add regression coverage for BOTH:

A. handoff:
- A terminal or departed;
- B explicitly admitted;
- B Running;
- UI shows B and controls B's exact turn.

B. existing multi-Agent semantics:
- Human explicitly addresses A and B in one Task;
- A and B both publish Running;
- Room retains BOTH authoritative executions;
- neither silently overwrites the other;
- a single-executor interrupt is rejected/hidden when control ownership is ambiguous (unless current server semantics already provide an exact owner).

The current executor-loss/composer fixes are otherwise good. Once server truth is no longer collapsed, this should be mergeable.

